PR #14, opened 2017-03-30
Converted all FD-ed benchmarking problems to complex step

hwangjt:


bouhlelma:
I think we should update the documentation of the toolbox. The current documentation does not match with the current version of the code anymore. We need to describe how we call a metamodel and options available in the toolbox (problems, sampling points etc.).

PR #62, opened 2017-09-14
Add a new benchmarking problem: branin function

relf:


relf:
Ok. I will modify my pull request with the result. Thanks.

PR #70, opened 2017-10-01
Install cython only if necessary

jschueller:


jschueller:
Ok, I changed my patch to install it only if necessary.

PR #75, opened 2017-10-04
changes to get ready to publish on pypi

hwangjt:


bouhlelma:
I think you need to modify the MANIFEST.in file

PR #82, opened 2017-11-14
Mixture of experts

relf:
This PR includes the following elements:
* the SMT extensions branch
* an initial version of mixture of experts (moe.py) 
* a new problem : L1 norm function
* a new script in examples dir: run_moe.py

I open this pull request to facilitate your feedback to get a first version of MOE in SMT.

PR #87, opened 2017-12-15
Ft mixture of experts

relf:
(Here again... see #82, #84 :relaxed:)
This PR includes the following elements:
* the SMT extensions branch
* an initial version of mixture of experts (moe.py)
* a new problem : Lp norm function
* test: extensions/tests/test_moe.py


relf:
Hmmm... Not sure to know why it passes on travis and not on appveyor. The error comes from a water_flow_lfidelity.py file in problems which still uses Problem instead of Problems.

relf:
My mistake, travis was certainly pending. I fixed Problems naming in two files. I didn't see these errors in my environment because I still had the problem.pyc file that made the tests pass!

bouhlelma:
the use of MOE in the unittest ('smt/smt/extension/tests' ) does not match the interface extensions: for example, the training points should be given during the initialization of MOE and the model should use the functions 'apply_method' and 'analysis_results' for training the model and making predictions.This would raise an error for the use of 'compute_rms_error' with extension (because extension does not implement the predict_values method), I suggest to add a key in 'compute_rms_error'  (key='surrogate' by default) which indicates if we are using 'surrogate' or 'extension'.

relf:
MOE implements surrogate interface even if it does not inherit from <code>SurrogateModel</code> class and for the time being I simply reused compute_rms_error as is. 
I feel that the use of the extension interface methods (apply and analyse_results) is a bit contrived. Actually, those methods bring no added value: it is like having a method <code>doit()</code>, the specifics are simply moved in the constructor. I think it would be better to have more focused interfaces such as <code>SurrogateExtension</code>, <code>OptimizerExtension</code> for extensions. But before that, unless you already foresee numerous extensions, we should accumulate extensions to see somehow if it is worth doing it.

PR #93, opened 2018-02-08
Add information in Readme.md + change the file name Readme.md to Surr…

bouhlelma:
…ogate Modeling Toolbox.md


relf:
Why renaming? A side-effect is that the landing page of the github repo does not display any information anymore, so I would suggest to keep README.md.

bouhlelma:
I agree!

PR #134, opened 2019-03-04
fixed GENN theory link in documentation

shb84:
Updated the link to GENN_theory.pdf in the documentation files and also removed some carry over metrics in neural_net/metrics.py that pertained to classification and were thus irrelevant for regression.


bouhlelma:
Is there a way to move the repo smt/smt/surrogate_models/neural_net towards smt/smt/utils/ in order to be more consistent with other surrogate modeling implementations?

relf:
Closed as it is included in #138

PR #135, opened 2019-03-08
deactivate plt.show when testflo is used && fix a bug in GEKPLS

bouhlelma:


relf:
Hi Mohamed. Not sure to understand the purpose of that change. Moreover there are still errors due to <code>plot_status</code>.

bouhlelma:
Hi Rémi, after installing the toolbox, we run the command 'testflo' to check if the toolbox runs correctly (as explained in the README). This run shouldn't open the figures generated from the test files. And we need to close all these figures manually in order to terminate the testflo run.

Another modification in the GEKPLS method, I found a typo in the computation of the PLS coefficients.

relf:
I've just pushed the PR #136 to fix the testflo/matplotlib interaction. In my windows environment, there was no problem, but indeed I tested on linux and have the windows popping up. The PR #136 fixes the matplotlib configuration (using no rendering backend Agg) in MOE, MFK tests. 

But I still got the windows poping up and by googling I found that calling switch_backend is necessary if matplotlib is already imported. I do not fully understand where matplotlib could be imported before, but it seems to work now. 

Maybe you could test #136, if it is ok merge and keep only the GEKPLS fix in that PR.

PR #138, opened 2019-04-12
bump version 0.3

relf:
This PR contains:
* the PR #134 + move neural_net dir under utils dir
* test against py37
* and then a version bump (0.3 as GENN was added)


bouhlelma:
It seems this PR responds to the comments of PR 134? Should we remove PR 134?

relf:
I've no problem with my Windows 7, py37 anaconda installation, but appveyor installation still fails due to numpy installation error (though I use same module version numbers)... I deactivate the test for the time being.

PR #158, opened 2019-07-11
Correlation bug fix in variance prediction

jbussemaker:
The correlation function was not correctly called in the variance prediction of Kriging-based surrogate models: the function call was done on the option value directly, which is a str, resulting in a TypeError exception.


jbussemaker:
For preventing this kind of stuff in the future, maybe it is a good idea to include unit tests.

relf:
Good catch! We definitely need more test coverage! Thanks.

PR #168, opened 2019-09-16
EGO optimizer

relf:
Add EGO optimizer as an SMT application.
(I've bumped version number to 0.4.0)


relf:
Thanks to Nathalie (@NatONERA) as well.
* The only reference is the Jones' paper. I've checked on my generated doc when you click on <code>[1]</code> it moves to the corresponding reference below. So... not sure what is wrong for you.
* I've bumped the version number in an attempt to follow [semantic versioning](https://semver.org/): 
** change from _extensions_ to _applications_ would have already required a major version number bump but as we are in 0.x version, API is not considered stable. Moreover we have not decided to tag a 1.0 (we could but it means no more API changes without warning/deprecation process). So a minor number bump is ok.
** adding a new feature like EGO requires to bump minor version number at least.
Hence the bump to 0.4.0. 0.3.5 would have only mean bug fix release.

bouhlelma:
- There are other references in the added text such as (Mockus, 1975), (Kushner etc.
- Understood, thanks.

relf:
Ok, I've misread your comment. I've focused on EGO part. I will add the other references.

PR #178, opened 2019-11-14
fix QP deriv computation for array outputs

fzahle:


fzahle:
hmm, seems like the Windows tests are now failing in places I didn't touch. any ideas?

relf:
Good catch! Thanks! 
Regarding KRG test failure, indeed I've tested on Windows and it fails repeatedly like with appveyor (with same exact values) while it works on Linux (!). It would suggest a difference out of SMT code? @bouhlelma, thoughts?
In the meantime, I would suggest to reactivate the test with a comment on the problem and a quick (and dirty?) adjustement of the tolerances (atol=0.2, rtol=0.03).

fzahle:
so I was lazy and only tested the models I was using, but it turned out the least squares model had the same issue, should also be fixed now.

relf:
> so I was lazy and only tested the models I was using...

Well, I think we have the same approach... Thanks for checking and fixing.
There is still a tolerance problem with KRG on MacOS... Just increase the tolerance (or may be just check the output dimension which is the point in the first place) so we can accept the PR.

Issue #180, opened 2019-11-28
QP predict_values not correct for multiple points


fzahle:
Hi,

It seems the QP model has a bug when predicting multiple points in one call:

    import numpy as np

    from smt.surrogate_models import QP, KRG
    from smt.examples.rans_crm_wing.rans_crm_wing import (
        get_rans_crm_wing,
        plot_rans_crm_wing,
    )

    xt, yt, xlimits = get_rans_crm_wing()

    interp = QP()
    interp.set_training_values(xt, yt)
    interp.train()
    v0 = np.zeros((4, 2))
    for ix, i in enumerate([10, 11, 12, 13]):
        v0[ix, :] = interp.predict_values(np.atleast_2d(xt[i, :]))
    v1 = interp.predict_values(np.atleast_2d(xt[10:14, :]))
    print('single nt', v0)
    print('multiple nt', v1)

    print('diff', v0 - v1) 

which produces the following outputs:

    single nt [[0.01004526 0.19395995]
     [0.01246052 0.34478044]
     [0.01835094 0.49012219]
     [0.02771651 0.6299852 ]]
    multiple nt [[0.01004526 0.01246052]
     [0.01835094 0.02771651]
     [0.19395995 0.34478044]
     [0.49012219 0.6299852 ]]
    diff [[-6.93889390e-18  1.81499430e-01]
     [-5.89041754e-03  3.17063930e-01]
     [-1.75609013e-01  1.45341748e-01]
     [-4.62405678e-01  0.00000000e+00]]

The issue popped up when using the VFM model, where the LF model is called with the full HF dataset in one call. works for KRG.

relf:
fixed by #181

PR #181, opened 2019-11-28
QP and VFM fixes

fzahle:
This is still work in progress.

A few fixes for VFD:

* VFM was quite slow, since every predict_values/predict_derivatives retrained the model.
* deriv computation failed for array outputs

lastly, I added a temporary fix for the QP issue with prediction at multiple sampling points. However, it turns out that the KRG+KRG results also differ with/without this fix. so there seems to be some issue there also.


relf:
Thank you for your contribution. I was able to push the QP multiple outputs fix in your pull request (fixes #180).

After some tolerance relaxation and refactoring for automated doc generation, it is ready for merging.

PR #211, opened 2020-07-09
Fixed errors in the variance computation in mfk.py

vindrouet:
- correction of sigma2, which is already descaled after _reduced_likelihood_function,
    and was descaled a second time in MFK
- correction of MSE line 480, should be Q_/(self.nt_all[i] - p - q)
    instead of Q_/(2*(self.nt_all[i] - p - q)) ; moreover, sigma2 = Q_/(self.nt_all[i] - p - q),
    so it is useless to compute Q_, and the expression is rewritten with sigma2 as in Le Gratiet and Cannamela 2015

Added a new test for the variance computation with mfk


smt/applications/mfk.py
```diff
@@ -462,10 +464,10 @@ def predict_variances_all_levels(self, X):
             beta = self.optimal_par[i]["beta"]
 
             # scaled predictor
-            sigma2 = self.optimal_par[i]["sigma2"]
+            sigma2 = self.optimal_par[i]["sigma2"] / self.y_std ** 2
             q = self.q_all[i]
-            p = self.p_all[i]
-            Q_ = (np.dot((yt - np.dot(Ft, beta)).T, yt - np.dot(Ft, beta)))[0, 0]
+            # p = self.p_all[i]
+            # Q_ = (np.dot((yt - np.dot(Ft, beta)).T, yt - np.dot(Ft, beta)))[0, 0]
```
relf, line 470: to be removed

PR #224, opened 2020-09-03
EGO refactoring

relf:
* remove doe from optimize returned values as already available in data points (doe is data[:ndoe, :])
* add doc comments
* add _setup_optimizer 
* protected method naming convention


relf:
started as a bug fix, finally ends up with some refactoring.

PR #229, opened 2020-09-14
Implement mixed_integer module which contains continuous relaxation logic

relf:
Following #228 , this PR contains:
* take continuous relaxation methods out of krg_based.py (vartypes option removal)
* create mixed_integer module which gathers previous methods
* methods renaming (if you have better proposal...)
  - <code>relax_limits</code> becomes <code>unfold_to_continuous_limits</code>
  - <code>project_values</code> becomes <code>cast_to_discrete_values</code>
  - <code>assign_labels</code> becomes <code>fold_with_enum_indexes</code>

To enforce consistency and provide support to the user, the mixed integer module exposes also 3 classes:
* <code>MixedIntegerSamplingMethod</code>: decorate SMT sampling methods to take into account mixed integer spec
* <code>MixedIntegerSurrogate</code>: decorate SMT surrogates which take into account mixed integer spec
* <code>MixedIntegerContext</code>: builder to ensure consistent usage of mixed integer, factory for previous classes


smt/applications/mixed_integer.py
```diff
@@ -0,0 +1,256 @@
...
+    return np.array(xlims)
+
+
+def cast_to_discrete_values(xtypes, x):
+    """
+    This function project continuously relaxed values to their closer assessable values.
+    Note: categorical (or enum) x dimensions are still expanded that is 
+    there are still as many columns as categorical possible values for the given x dimension. 
+    For instance, if an input dimension is typed ["blue", "red", "green"] in xlimits a sample/row of 
+    the input x may contain the values (or mask) [..., 0, 0, 1, ...] to specuify "green" for
```
Paul-Saves, line 110: specuify

smt/applications/mixed_integer.py
```diff
@@ -0,0 +1,256 @@
...
+            raise ValueError(
+                "Bad var type specification: "
+                "should be 'cont', 'int' or ('cate', n), got {}".format(xtyp)
+            )
+    return res
+
+
+def unfold_to_continuous_limits(xtypes, xlimits):
+    """
+    This function unfold xlimits to add continuous dimensions
```
Paul-Saves, line 66: unfolds

smt/applications/mixed_integer.py
```diff
@@ -0,0 +1,256 @@
...
+            raise ValueError(
+                "Bad var type specification: "
+                "should be 'cont', 'int' or ('cate', n), got {}".format(xtyp)
+            )
+    return ret
+
+
+def fold_with_enum_indexes(xtypes, x):
+    """
+    This function reduce categorical inputs from discrete unfold space to
```
Paul-Saves, line 153: unfolds

smt/applications/mixed_integer.py
```diff
@@ -0,0 +1,256 @@
...
+    res = 0
+    for xtyp in xtypes:
+        if xtyp == FLOAT or xtyp == INT:
+            res += 1
+        elif isinstance(xtyp, tuple) and xtyp[0] == ENUM:
+            res += xtyp[1]
+        else:
+            raise ValueError(
+                "Bad var type specification: "
+                "should be 'cont', 'int' or ('cate', n), got {}".format(xtyp)
```
Paul-Saves, line 59: I don't understand if the user should write as before (int, cont, cate) or INT, FLOAT, ENUM

Paul-Saves (changes requested):
The error messages says that xtype should be written as ["cont", "int", ("cate",2)] but the tests are written as [FLOAT, INT, (ENUM,2)]

smt/applications/mixed_integer.py
relf: I've defined INT="int", FLOAT="cont" and ENUM="cate" variables in mixed_integer (I think float and enum are more common than cont and cate). I've to change the error messages. The idea is to use constant variables instead of strings because if you make a typo (ex. FLOT) you get an error message (otherwise the string "flot" will pass).

Paul-Saves: Ok, then, it works with either "cont" or FLOAT. Yes, I understand your point of view, these are more common name in computer science/programming.

Paul-Saves (approved):

PR #237, opened 2020-10-14
Mixed surrogate model

raul-rufato:
I am a student at ISAE-Supaero and I do research with Professor Morlier. I am implementing the Gower distance for kernel calculation of mixed variables (only available for KRG). I don't know how I should go about creating the documentation for this pull request, as I just added it as an extra case inside MixedIntegerSurrogate. I added a notebook in the home folder that performs an example using the Gower distance.

Thank you. 

Regards,

Raul


smt/surrogate_models/krg_based.py
```diff
@@ -65,19 +69,49 @@ def _new_train(self):
...
+                if not isinstance(Xt, np.ndarray): 
+                    is_number = np.vectorize(lambda x: not np.issubdtype(x, np.number))
+                    cat_features = is_number(Xt.dtypes)    
+                else:
+                    cat_features = np.zeros(x_n_cols, dtype=bool)
+                    for col in range(x_n_cols):
+                        if not np.issubdtype(type(Xt[0, col]), np.number):
+                            cat_features[col]=True   
+                
+                    if not isinstance(Xt, np.ndarray): Xt = np.asarray(Xt)
```
relf, line 89: already tested above?

smt/surrogate_models/krg_based.py
```diff
@@ -338,24 +405,53 @@ def _predict_variances(self, x):
...
             - self._regression_types[self.options["poly"]](x).T,
         )
+        else:
+            # Compute the correlation function
+            Xt = x
+            x_n_rows, x_n_cols = Xt.shape
+            r = np.exp(-gower_matrix(x,data_y=self.X_train,weight = np.asarray(self.optimal_theta)))
+            n_samples, n_features = x.shape
+                
+            if not isinstance(Xt, np.ndarray):
```
relf, line 432: same code as in initialize? Is it useful? If yes should be in a separate function

smt/surrogate_models/krg_based.py
```diff
@@ -258,21 +293,49 @@ def _predict_values(self, x):
...
-            self.optimal_theta, d
-        ).reshape(n_eval, self.nt)
-        y = np.zeros(n_eval)
-        # Compute the regression function
-        f = self._regression_types[self.options["poly"]](x)
-        # Scaled predictor
-        y_ = np.dot(f, self.optimal_par["beta"]) + np.dot(r, self.optimal_par["gamma"])
-        # Predictor
-        y = (self.y_mean + self.y_std * y_).ravel()
+        if self.options["corr"] != "gower_corr":
```
relf, line 296: better to test for equality

relf (changes requested):
Binary files utils/neural_net/__pycache__/... should not be included in the PR.
You should add some tests in tests/test_krg.py to check proper results of your code
We also use the code formatter named black, you should use it on the file you modify.
Finally your code seems to break some other tests in SMT, so you have to ensure that the SMT test suite pass at least in your environment.

Paul-Saves:
You should merge from SMTorg/SMT master branch. My branch is not up to date anymore. 
Moreover, mixed integer were implemented in the master. 
So, when you merge, you should pull master branch.

PR #239, opened 2020-11-02
Add notebook documentation

Paul-Saves:


relf (changes requested):
Nitpicking: Would you mind renaming in SMT_MixedInteger_application.ipynb to have some kind of naming consistency?

Paul-Saves:
See Slack. MGP model option has not be enable in the applications and decorators.

smt/applications/ego.py
```diff
@@ -17,7 +17,7 @@
 from smt.applications.mixed_integer import MixedIntegerContext
 from smt.utils.misc import compute_rms_error
 
-from smt.surrogate_models import KPLS, KRG, KPLSK
+from smt.surrogate_models import KPLS, KRG, KPLSK,MGP
```
relf, line 20: Nitpicking: would you mind passing black or adding a space after the comma?

smt/applications/ego.py
```diff
@@ -107,7 +107,7 @@ def _initialize(self):
         declare(
             "surrogate",
             KRG(print_global=False),
-            types=(KRG, KPLS, KPLSK),
+            types=(KRG, KPLS, KPLSK,MGP),
```
relf, line 110: same as previous

relf (changes requested):

smt/applications/ego.py
Paul-Saves: I applied black on the complete folder so other files as marked as changed

relf: Yes, I've seen that. It is ok.

relf (approved):

PR #242, opened 2020-11-03
Check kernel compatibility (KPLS, KPLSK) for different correlations

NatOnera:
add KPLS example in run_examples.py
add message error for KPLS + matern kernels
add message error for KPLSK + abs + matern kernels
try to modify initialization of KPLS and KPLSK to update automatically documentation


smt/examples/run_examples.py
```diff
@@ -270,6 +270,22 @@
 if plot_status:
     plt.show()
 
+#KPLS + absolute exponential correlation kernel
+# The variables 'name' must be equal to 'KPLS'. 'n_comp' and 'theta0' must be
+# an integer in [1,ndim[ and a list of length n_comp, respectively. Here is an
```
relf, line 275: should be [1, ndim]

relf (changes requested):

smt/examples/run_examples.py
relf: ok you can have n_comp = ndim, but in that case KPLS is useless

relf (approved):

PR #248, opened 2020-12-01
Fix kriging reduced likelihood computation

relf:
This PR fixes the classc kriging (KRG) computation (regression due to MGP integration) but gradient still have to be adjusted accordingly.


relf:
Actually previous kriging example result was not correct (not converged). SMT 0.7.0 fixed that and it shows a different result.

PR #252, opened 2020-12-04
changing bounds theta, allowing different parameters per lvl MFK

anfelopera:


relf (changes requested):

smt/applications/mfk.py
```diff
@@ -208,10 +208,32 @@ def _new_train(self):
         self.optimal_rlf_value = nlevel * [0]
         self.optimal_par = nlevel * [{}]
         self.optimal_theta = nlevel * [0]
-        self.X_norma_all = [(x - self.X_mean) / self.X_std for x in X]
+        self.X_norma_all = [(x - self.X_offset) / self.X_scale for x in X]
         self.y_norma_all = [(f - self.y_mean) / self.y_std for f in y]
-
+        
+        #### added by A. Lopez-Lopera, 27/11/2020 ####
```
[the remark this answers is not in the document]
anfelopera: ok, removing notes

anfelopera (commented):
removing notes about modifications and using is_unit_X instead

smt/utils/kriging_utils.py
```diff
@@ -50,37 +57,44 @@ def standardization(X, y, copy=False):
...
+            min if isUnitX=True) of each input variable.
 
     y_std:  list(1)
             The standard deviation of the output variable.
 
     """
-    X_mean = np.mean(X, axis=0)
-    X_std = X.std(axis=0, ddof=1)
+    
+    if isUnitX:
```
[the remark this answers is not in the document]
anfelopera: Done.

smt/utils/kriging_utils.py
```diff
@@ -20,11 +20,13 @@
...
 
 
-def standardization(X, y, copy=False):
+def standardization(X, y, copy=False, is_Unit_X=False):
 
     """
+
     We substract the mean from each variable. Then, we divide the values of each
-    variable by its standard deviation.
+    variable by its standard deviation. If isUnitX, we scale the input space X to
```
relf, line 28: Should be <code>is_unit_X</code> but I wonder whether it is the right name. When you call the method you ask for an action (like copy), starting with an action verb would be better, may be something like <code>scale_X_to_unit</code>?

smt/utils/kriging_utils.py
```diff
@@ -20,11 +20,13 @@
 # TODO: Create hyperclass Kernels and a class for each kernel
 
 
-def standardization(X, y, copy=False):
+def standardization(X, y, copy=False, scale_X_to_unit=False):
```
relf, line 23: I've just noticed we never set copy or scale_X_to_unit options to True, so we just have to remove them.

relf (changes requested):

smt/utils/kriging_utils.py
anfelopera: I agree with copy (not sure when it will True) but scale_X_to_unit may interact with future developments. I would prefer to keep it.

PR #257, opened 2020-12-15
to enable defining bounds for theta and noise, the value of the nugget and a noise variance per design input

anfelopera:


doc/_src_docs/surrogate_models/krg.rst
```diff
@@ -12,17 +12,34 @@ Kriging is an interpolating model that is a linear combination of a known functi
...
+  
+.. math ::
+  \left(1 + \sqrt{5}\sum\limits_{l=1}^{nx}\theta_{l}\left|x_l^{(i)}-x_l^{(j)}\right| + \frac{5}{3}\sum\limits_{l=1}^{nx}\theta_{l}^{2}\left(x_l^{(i)}-x_l^{(j)}\right)^{2}\right) \exp\left(-\sqrt{5}\sum\limits_{l=1}^{nx}\theta_{l}\left|x_l^{(i)}-x_l^{(j)}\right|\right), \\
+  \text{Mat\'ern 5/2 corralation function}
+  
+.. math ::
+  \left(1 + \sqrt{3}\sum\limits_{l=1}^{nx}\theta_{l}\left|x_l^{(i)}-x_l^{(j)}\right|\right) \exp\left(-\sqrt{3}\sum\limits_{l=1}^{nx}\theta_{l}\left|x_l^{(i)}-x_l^{(j)}\right|\right).\\
+  \text{Mat\'ern 3/2 corralation function}
 
 These two correlation functions are called by 'abs_exp' (exponential) and 'squar_exp' (Gaussian) in SMT.
```
Paul-Saves, line 33: _" These two correlation functions are called by 'abs_exp' (exponential) and 'squar_exp' (Gaussian) in SMT. "_
 
 Why "these two" ? There is 4 kernels.
 What are the abbreviations for Matérn ?

smt/surrogate_models/krg_based.py
```diff
@@ -75,7 +86,23 @@ def _initialize(self):
...
+            desc="Initial noise hyperparameter",
+        )
+        declare(  # I let the bounds previously imposed by default (how were they fixed?)
+            "noise_bounds",
+            [1e-16, 1e10],
+            types=(list, np.ndarray),
+            desc="bounds for hyperparameters",
+        )
+        declare(
+            "is_noise_het",
```
Paul-Saves, line 101: Use_heteroscedastic_noise ?

smt/surrogate_models/krg_based.py
```diff
@@ -130,7 +180,9 @@ def _new_train(self):
             self._specific_train()
         else:
             if self.options["eval_noise"]:
-                self.optimal_theta = self.optimal_theta[:-1]
+                if not self.options["is_noise_het"]:
+                    self.noise = self.optimal_theta[self.D.shape[1] :]
```
Paul-Saves, line 184: Why the case heteroscedastic is not treated like the homoscedastic one ?
Can't we uniformize these two ?
Just by adding the length ?

smt/surrogate_models/krg_based.py
```diff
@@ -57,9 +55,22 @@ def _initialize(self):
             desc="Correlation function type",
             types=(str),
         )
+        declare(
+            "nugget", [], types=(float, list), desc="a jitter for numerical stability"
```
Paul-Saves, line 59: Default value missing.

10.0 * np.finfo(np.double).eps

smt/surrogate_models/krg_based.py
```diff
@@ -187,24 +234,27 @@ def _reduced_likelihood_function(self, theta):
...
-            if self._lvl != self.nlvl:
-                # in the case of multi-fidelity optimization
-                # it is very probable that lower-fidelity correlation matrix
-                # becomes ill-conditionned
-                nugget = 10.0 * nugget
-        elif self.name in ["MGP"]:
-            nugget = 100.0 * nugget
+        nugget = self.options["nugget"]
+        if self.options["eval_noise"]:
+            nugget = 0
```
Paul-Saves, line 239: Why nugget= 0 ? We want to allow both nugget and noise usage.

smt/surrogate_models/krg_based.py
```diff
@@ -187,24 +234,27 @@ def _reduced_likelihood_function(self, theta):
...
+        else:
+            if not self.options["nugget"]:
+                MACHINE_EPSILON = np.finfo(np.double).eps
+                nugget = 1e2 * MACHINE_EPSILON
+
```
Paul-Saves, line 244: To remove

smt/surrogate_models/krg_based.py
```diff
@@ -187,24 +234,27 @@ def _reduced_likelihood_function(self, theta):
...
+        elif self.options["eval_noise"] and not self.options["is_noise_het"]:
+            theta = tmp_var[0 : self.D.shape[1]]
+            noise = tmp_var[self.D.shape[1] :]
         r = self._correlation_types[self.options["corr"]](theta, self.D).reshape(-1, 1)
 
-        R = np.eye(self.nt) * (1.0 + nugget + noise)
+        if self.options["is_noise_het"]:
+            R = np.eye(self.nt) * (1.0 + nugget + noise / self.nt_reps)
+        else:
+            R = np.eye(self.nt) * (1.0 + nugget + noise)
```
Paul-Saves, line 257: As before, homoscedastic is just heteroscedastic with only one value.
Please avoid if else structure

smt/surrogate_models/krg_based.py
```diff
@@ -253,6 +306,9 @@ def _reduced_likelihood_function(self, theta):
         reduced_likelihood_function_value = -(self.nt - p - q) * np.log10(
             sigma2.sum()
         ) - self.nt * np.log10(detR)
+        # if self.options["is_noise_het"]:
+        #     lambdas = (self.nt_reps-1) * noise + np.log10(self.nt_reps)
+        #     reduced_likelihood_function_value += lambdas.sum()
```
Paul-Saves, line 311: To be removed

smt/surrogate_models/krg_based.py
```diff
@@ -833,18 +883,28 @@ def grad_minus_reduced_likelihood_function(log10t):
             k, incr, stop, best_optimal_rlf_value, max_retry = 0, 0, 1, -1e20, 10
             while k < stop:
                 # Use specified starting point as first guess
-                if self.options["eval_noise"]:
+                if self.options["eval_noise"] and not self.options["is_noise_het"]:
```
Paul-Saves, line 886: Same remark

Paul-Saves (changes requested):

Paul-Saves:
Lack of examples  with heteroscedastic noise. 
Lack of documentation with heteroscedastic noise.

doc/_src_docs/surrogate_models/krg.rst
anfelopera: I did see that line but it is already corrected!

smt/surrogate_models/krg_based.py
anfelopera: It is defined later if eval_noise = True

smt/surrogate_models/krg_based.py
anfelopera: There is no need if there is a noise... The noise will play the role of the nugget

smt/surrogate_models/krg_based.py
anfelopera: Work in progress!

smt/surrogate_models/krg_based.py
Paul-Saves: The nugget is for numerical problems. it is homoscedastic and fixed.
The noise is for noisy data, it could be both homoscedastic or heteroscedastic.

Evaluating a noise to play the role of the nugget is dirty. It could word but it is not a good way to do it.

Paul-Saves: Eval_noise should evaluate the noise, not the sum "nugget+noise" sometimes and just "noise" otherwise

anfelopera: No, we are not going to use the noise as a nugget... The nugget has to be considered only in cases with numerical problems... For the noisy case, we do not need that nugget because there is no such problem because of the influence of the noise.

Paul-Saves: Can you at least, set the default nugget to 0 and remove the "set to 0" with eval_noise.

Paul-Saves: - "nugget", 0, types=(float, list), desc="a jitter for numerical stability" 
.....
- if not self.options["eval_noise"]:
                MACHINE_EPSILON = np.finfo(np.double).eps
                nugget = 1e2 * MACHINE_EPSILON

Paul-Saves: Another question, why "list" type for nugget ?

smt/surrogate_models/krg_based.py
anfelopera: In this initial implementation, the noise variances are pointwise sensible estimates via (Ankenman et al., 2010), and they are not estimated via ML, i.e. they are fixed for the MLE. In future developments, they will be integrated in the MLE.

anfelopera:
> --
> 
> * Lack of examples  with heteroscedastic noise.
> * Lack of documentation with heteroscedastic noise.
> 
> Coverage has decreased.

smt/surrogate_models/krg_based.py
anfelopera: We can do it but that forces the initialisation of the noise variances in a list, p.e. : noise0 = nt * [1e-6] ... and check the option using len(noise).

PR #265, opened 2021-01-08
Added convenience methods

LDAP:
- Make SamplingMethod a Generator for use in for-loops
- Make SurrogateModel callable for intuitive use as function


LDAP:
The CI failing should not be related to the changes

relf:
* Sampling methods generate points all at once, there is no meaning in calling compute for one point.  
* Regarding the surrogate model callable API, unless there is a strong argument for it, I prefer the explicit <code>predict_values</code> call and do not want to have several ways for doing the same thing.

LDAP:
> 
> 
>     * Sampling methods generate points all at once, there is no meaning in calling compute for one point.
> 
>     * Regarding the surrogate model callable API, unless there is a strong argument for it, I prefer the explicit `predict_values` call and do not want to have several ways for doing the same thing.

The send method does provide special features to the class in python. See the docs [here](https://docs.python.org/3/reference/expressions.html#generator.send)

The reason is not to call compute for one point, instead it enables code like this:
```python
...
sm = Random(xlimits=xlimits)
# Iterator 
for sample in sm:
    ...
# or
transformed_samples = [transform(sample) for sample in sm]
# or use as generator
transformed_sample_generator = (transform(sample) for sample in sm)
```

Also, it enables coroutines see https://www.python.org/dev/peps/pep-0342/

Using generators is considered as good practice ([as described here](https://www.freecodecamp.org/news/how-and-why-you-should-use-python-generators-f6fb56650888/)). 
The Sampling Methods could also overwrite the send method and provide a more memory/computationally efficient generator. Also it enables Sampling Methods to receive the parameter value, allowing them to generate context dependent samples. This is part of the coroutine pattern.

relf:
I do not contest the generator pattern, but taking into account your implementation, the code below (your example) just loops for ever.

```python
import numpy as np
from smt.sampling_methods import Random

xlimits = np.array([[0.0, 1.0]])
sm = Random(xlimits=xlimits)
for x in sm:
    print(x)
```

LDAP:
> 
> 
> I do not contest the generator pattern, but taking into account your implementation, the code below (your example) just loops for ever.
> 
> ```python
> import numpy as np
> from smt.sampling_methods import Random
> 
> xlimits = np.array([[0.0, 1.0]])
> sm = Random(xlimits=xlimits)
> for x in sm:
>     print(x)
> ```

This is expected, as this represents a stream (you can find something like this in Java as Streams or in Haskell as infinite lists). You would instead use the sample to for example update your model and when your model is good enough or fulfills a certain criteria you would break the loop.

PR #266, opened 2021-01-09
Fix IndexError in predict_output_derivatives and predict_variances

LDAP:
`predict_output_derivatives` and `predict_variances` hat missing `check_2d_array` calls which ensure that the numpy array is converted to a 2d array.
To prevent that again, check_2d_array was refactored to ensure_2d_array, to reflect that the function does break the 
Command-Query-Responsibility-Segregation and does not return the same array in case the dimension is not 2.


Paul-Saves:
See #262

PR #267, opened 2021-01-09
Added type_hints to public und developer API

LDAP:
Numpy (>=1.20.0) started to ship with type hints included. I started adding type hints to the public and developer APIs. 

Some may still be too restrictive or to unspecific.


LDAP:
The CI failing should not be related to the changes

relf:
> Some may still be too restrictive or to unspecific.

Looks good to me since Github Actions is happy. What "restrictive or unspecific" point do you have in mind ?

LDAP:
> 
> 
> > Some may still be too restrictive or to unspecific.
> 
> Looks good to me since Github Actions is happy. What "restrictive or unspecific" point do you have in mind ?

Numpy does for example often allow input values of type array_like. But as far as I know they do not expose a python definition of the type (may be added in numpy 1.20). This can be revisited when numpy 1.20 releases.

relf:
I am not familiar with type hinting. To be sure, I did not miss something: Do you mean that the type hinting in SMT here would break depending on some choices made by numpy?  :eyes: or just that the type hinting could be refined later on following some guidelines of numpy.

LDAP:
> 
> 
> I am not familiar with type hinting. To be sure, I did not miss something: Do you mean that the type hinting in SMT here would break depending on some choices made by numpy? 👀 or just that the type hinting could be refined later on following some guidelines of numpy.

The latter. Moreover, type hints are ignored by the python runtime, they allow IDEs and static codecheckers better type hinting, leading to better code quality and fewer runtime type errors.

relf (approved):

PR #268, opened 2021-01-11
Adapt predict_variance_derivatives to SMT codestyle

LDAP:


Paul-Saves (approved):
Seems great. My thanks to you.

smt/surrogate_models/surrogate_model.py
```diff
@@ -384,15 +385,17 @@ def predict_variances(self, x):
...
+
         Returns:
         --------
-        - derived_variance: array_like
-        The jacobian of the variance of the kriging model
+        derived_variance: np.ndarray
+            The jacobian of the variance
         """
         check_support(self, "derivatives")
         check_support(self, "variances")
```
relf, line 401: Much better thanks! Then you can replace the two lines above by <code>check_support(self, "variance_derivatives")</code>

relf (changes requested):

smt/surrogate_models/surrogate_model.py
LDAP: Should be enough to delete them since check_support(...) is called in _predict_variance_derivatives?

LDAP: I am going to add it to match the other methods.

relf (approved):

Issue #272, opened 2021-01-11
Tests (randomly) failing


LDAP:
- GitHub CI fails randomly 
- I had locally never a run with all tests passed, [pytest.log](https://github.com/SMTorg/smt/files/5798046/pytest.log), commit: 9f6610ef31b37cc2a1bfaec7fe1a0f24610f550d
I am using Windows 10, Python 3.8.3.

Paul-Saves:
Have you tried it after #269 and #270  ? (merge today)

relf:
I reproduce this kind of errors with scikit-learn 0.24. It works fine after downgrading to sklearn 0.23.2 (on Windows 7, python 3.7). Could you try?

LDAP:
I can confirm: All tests passing with scikit-learn 0.23.2.
Is this an upstream bug in scikit-learn or fixes a 0.24 a bug in 0.23.2 which led to false results?

In any case I would prefer to enforce the dependency version ether via setup.py install_requires or at least to add a requirements.txt to the repo, that the developers who mostly run the tests do have a consistent development environment (This file could also define the dependencies to run the setup (numpy and Cython)).

relf:
> Is this an upstream bug in scikit-learn or fixes a 0.24 a bug in 0.23.2 which led to false results?

We just discover the problem... To be investigated...

> In any case I would prefer to enforce the dependency version ether via setup.py install_requires or at least to add a requirements.txt to the repo, that the developers who mostly run the tests do have a consistent development environment (This file could also define the dependencies to run the setup (numpy and Cython)).

I am a bit reluctant to fix a required version. If we had here we would have discovered the problem maybe much later and scikit-learn 0.24 seems to be ok for surrogate models without PLS, so maybe a requirements.txt... and you're right <code>setup.py</code> has to be updated.

relf:
superseded by #274

PR #273, opened 2021-01-12
Fix setup

LDAP:
- Autoinstall dependencies needed for setup.py (numpy, Cython, wheel, setuptools)
- Added requirements.txt for developers

Should I add a specific version for scikit-learn as mentioned in #272?

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -15,6 +15,7 @@ SMT has been developed thanks to contributions from:
 * Frederick Zahle <frza@dtu.dk>
 * Jasper Bussmaker <jasper.bussemaker@gmail.com>
 * Julien Schueller <schueller@phimeca.com>
+* Lucas Alber <lucasd.alber@gmail.com>
 * Mostafa Meliani <melimostafa@gmail.com>
 * Nina Moëllo <nina.moello@gmail.com>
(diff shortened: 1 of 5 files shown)


LDAP:
[PEP 518](https://www.python.org/dev/peps/pep-0518/)
[Cython Docs](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#basic-setup-py)

relf:
Thanks for the references and you can also add your name in the AUTHORS.md file.

relf:
> Should I add a specific version for scikit-learn as mentioned in #272?

As the errors (#272 ) seems to be related to surrogate models not well adapted to test functions I would prefer not to mention a specific version besides 0.24 works well for some other surrogate models.

AUTHORS.md
```diff
@@ -23,3 +23,4 @@ SMT has been developed thanks to contributions from:
 * Rémy Priem <remy.priem@onera.fr>
 * Steven Berguin <stevenberguin@gatech.edu>
 * Vincent Drouet <vincent.drouet@onera.fr>
+* Lucas Alber <lucasd.alber@gmail.com>
```
relf, line 26: Well so far we have used the lexicographic order  :wink:

relf (approved):

PR #282, opened 2021-01-20
MOE Variance Prediction and SurrogateModel

jbussemaker:
This pull request contains two small updates to the Mixture of Expert features: implementation of variance prediction and wrapping the MOE application in a `SurrogateModel` subclass.

# Variance Prediction

- For smooth recombination, the variance of the MOE is the sum of the variances times the square of the probabilities (according to [Bartoli et al., 2016](https://www.researchgate.net/publication/303902935_Improvement_of_efficient_global_optimization_with_application_to_aircraft_wing_design), Eq. 15)
- For hard recombination, the variance prediction of the chosen expert is simply taken

# Wrapping as a `SurrogateModel` Class

The MOE class is wrapped by the `MOESurrogateModel` class, to expose the MOE features through the `SurrogateModel` API. Options are automatically shared.


relf (approved):
Hi Jasper, thanks for your contribution on the variance prediction and for shortening the TODO list! I like the automated sharing of options between MOE and MOESurrogateModel. Do you want to add anything before merging?

jbussemaker:
Hi Rémi, no I think everything is fine as it is. If it could be released in a new (minor) update version on PyPI soon that'd be great, then I can start using these features in other code I'm currently writing :)

relf:
As 0.8 was just released, I would prefer to wait a bit for a 0.9 and add some documentation about MOE. In the meantime, you can install from Github like this :
```
> pip install git+https://github.com/SMTorg/smt.git@c904913 
```

PR #284, opened 2021-01-27
Do not scale the samples by default

LDAP:
The \_\_call\_\_ method does scale the result from _compute() by default. This is unexpected and not documented. Also, it does prevent subclasses to generate samples in the input space directly.

### DIFF
--- a/smt/sampling_methods/full_factorial.py
+++ b/smt/sampling_methods/full_factorial.py
@@ -5,6 +5,7 @@ This package is distributed under New BSD license.
 
 Full-factorial sampling.
 """
+from smt.utils.misc import scale_to_xlimits
 import numpy as np
 
 from smt.sampling_methods.sampling_method import SamplingMethod
@@ -63,4 +64,4 @@ class FullFactorial(SamplingMethod):
         for kx in range(nx):
             x[:, kx] = x_list[kx].reshape(np.prod(num_list))[:nt]
 
-        return x
+        return scale_to_xlimits(x, xlimits)
--- a/smt/sampling_methods/full_factorial.py
+++ b/smt/sampling_methods/full_factorial.py
@@ -5,13 +5,12 @@ This package is distributed under New BSD license.
 
 Full-factorial sampling.
 """
-from smt.utils.misc import scale_to_xlimits
 import numpy as np
 
-from smt.sampling_methods.sampling_method import SamplingMethod
+from smt.sampling_methods.sampling_method import ScaledSamplingMethod
 
 
-class FullFactorial(SamplingMethod):
+class FullFactorial(ScaledSamplingMethod):
     def _initialize(self):
         self.options.declare(
             "weights",
@@ -28,7 +27,9 @@ class FullFactorial(SamplingMethod):
 
     def _compute(self, nt):
         """
-        Compute the requested number of sampling points.
+        Implemented by sampling methods to compute the requested number of sampling points.
+
+        The number of dimensions (nx) is determined based on `xlimits.shape[0]`.
 
         Arguments
         ---------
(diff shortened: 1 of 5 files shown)


relf:
I agree we should document the fact the `__call__` method does a scaling and that `_compute` is expected to give samples in the [0., 1.] hypercube. At the moment, all sampling methods are intended to follow that pattern that we recently enforced by adding the @abstractmethod on `_compute`. :wink:

If we imagine a sampling method not following that pattern, maybe it means that the `@abstractmethod` should be moved on the `__call__` method which is actually the true API to respect or we just have to remove the `@abstractmethod` as I feel now it is just getting in the way.

LDAP:
At least I would ensure that sampling methods are able to compute samples outside the [0,1] hypercube. Also I find it confusing that subclasses should overwrite a \_\_call\_\_ method which does have a special meaning in python.

In any case, I would prefer the call to a utility method instead of using the template method pattern. (here: scale_to_xlimits)

relf:
> At least I would ensure that sampling methods are able to compute samples outside the [0,1] hypercube

Not sure to understand here: do you mean we could ensure that the `_compute` return only samples in [0., 1.] hypercube?

I prefer to wait for an actual sampling method generating directly in the input space before duplicating `scale_to_xlimits` calls everywhere or moving the `@abstractmethod`. As for now, I find the current implementation more straightforward.

Though a bit awkward for sure, the override of `__call__` with a no-op implementation for `_compute` is still possible for such new sampling method. And in in this case, we could reconsider extracting `scale_to_xlimits`.

LDAP:
> 
> 
> > At least I would ensure that sampling methods are able to compute samples outside the [0,1] hypercube
> 
> Not sure to understand here: do you mean we could ensure that the `_compute` return only samples in [0., 1.] hypercube?

No, I would allow a sampling method to do so.

> I prefer to wait for an actual sampling method generating directly in the input space before duplicating `scale_to_xlimits` calls everywhere or moving the `@abstractmethod`. As for now, I find the current implementation more straightforward.

Actually, that is why I proposed this pull request. I am currently working on a sampling method which does create samples in the input space directly.

relf:
... or remove the `@abstractmethod` enforcement (so you can just override `__call__`) letting options open for the developper, make `scale_to_xlimits` a method, keep `_compute` and the template pattern method like:

```python
def __call__(self, n):
    x = self._compute(n)
    return self._scale_to_xlimits(x)
```

LDAP:
> 
> 
> ... or remove the `@abstractmethod` enforcement (so you can just override `__call__`) letting options open for the developper, make `scale_to_xlimits` a method, keep `_compute` and the template pattern method like:
> 
> ```python
> def __call__(self, n):
>     x = self._compute(n)
>     return self._scale_to_xlimits(x)
> ```

Personally I find this very dangerous: A developer then has to know that overwriting `_compute` does scale the samples while overwriting `__call__` would be the correct option. Since the util module is also public I would suggest the developer to use `scale_to_xlimits` directly.

The `@abstractmethod` annotation is in this case a good thing because a developer knows directly which method he must overwrite. We could add a hint to scale_to_xlimits in the docstring so that the developer does not need to rewrite existing code.

relf:
In any case I agree the documentation has to be updated. Still I am not fan of the scale_to_xlimits call duplication and a separate utility very specific to sampling methods.

Let's introduce an intermediate class, `ScaledSamplingMethod`?, which inherits from a `SamplingMethod` which implements `__call__` with the template method pattern above and exposes the `_compute` (with proper docstring) to be implemented. Current sampling methods will then inherit from ScaledSamplingMethod and you can implement yours inheriting directly from `SamplingMethod`.

LDAP:
I do like this approach. How about `UnitIntervalSamplingMethod` to make clear that the samples have to be generated in the unit interval.

relf:
Yes I have tought about `UnitIntervalSamplingMethod` but then found a bit weird that `__call__` returns something not in the unit interval, so I would keep `ScaledSamplingMethod` and rather document the class and the `_compute` method properly.

LDAP:
See 0581492

relf (approved):
Excellent!

PR #286, opened 2021-01-31
Gower distance based Mixed Surrogate Model

raul-rufato:
I'm Raul Carreira Rufato, I'm implementing Gower's distance to calculate a mixed kernel (continuous and categorical variables).


raul-rufato:
Some errors occurd, during the verification

PR #293, opened 2021-02-12
add a small multistart

Paul-Saves:
add a small multistart to avoid having too often really bad hyperparameters.

### DIFF
--- a/smt/surrogate_models/krg_based.py
+++ b/smt/surrogate_models/krg_based.py
@@ -212,6 +212,7 @@ class KrgBased(SurrogateModel):
             self.optimal_theta,
         ) = self._optimize_hyperparam(D)
         
+        
         if self.name in ["MGP"]:
             self._specific_train()
         else:
@@ -1143,7 +1144,6 @@ class KrgBased(SurrogateModel):
                                     optimal_theta_res= optimal_theta_res_loop
                         
                     elif self.options["hyper_opt"] == "TNC":
-
                         optimal_theta_res = optimize.minimize(
                             minus_reduced_likelihood_function,
                             theta0,
@@ -1161,10 +1161,9 @@ class KrgBased(SurrogateModel):
                             bounds=bounds_hyp,
                             options={"maxiter": 100},
                         )
-                        
-                        theta0_loop= np.ones(theta0.shape)*-6
+                        theta0_loop= 10**(np.ones(theta0.shape)*-6)
                         for i in range(7):
-                            theta0_loop=theta0_loop+np.ones(theta0.shape)
+                            theta0_loop=10**(np.log10(theta0_loop)+np.ones(theta0.shape))
                             optimal_theta_res_loop = optimize.minimize(
                             minus_reduced_likelihood_function,
                             theta0_loop,
@@ -1176,12 +1175,11 @@ class KrgBased(SurrogateModel):
                             if optimal_theta_res_loop["fun"]< optimal_theta_res["fun"] : 
                                 optimal_theta_res= optimal_theta_res_loop
                                 optimal_theta_res_2 = optimal_theta_res_loop
-                                
                         if self.options["eval_noise"] and not self.options["use_het_noise"] : 
-                            theta0_loop= -12*np.ones(theta0.shape)
+                            theta0_loop= 10**(-12*np.ones(theta0.shape))
                             theta0_loop[:len(theta0_loop)-len(self.noise0)]=optimal_theta_res["x"][:len(theta0_loop)-len(self.noise0)]
                             for i in range(5):
-                                theta0_loop=theta0_loop+3*np.ones(theta0.shape)
+                                theta0_loop=10**(np.log10(theta0_loop)+3*np.ones(theta0.shape))
                                 theta0_loop[:len(theta0_loop)-len(self.noise0)]=optimal_theta_res["x"][:len(theta0_loop)-len(self.noise0)]
 
                                 optimal_theta_res_loop = optimize.minimize(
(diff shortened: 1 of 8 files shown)


Paul-Saves:
We normalize data before computing the hyperparameters, which are "auto-correlation length". 
For the optimization of the hyperparameters, we use scipy.optimize COBYLA and it performs a local optimization.

I tried to add a small multistart, it is not perfect but better than before. You find a real problem. It is not normal that one should set the starting point every time.

I used your example : 
Before : 
image

After : 
image

smt/surrogate_models/krg_based.py
```diff
@@ -1082,18 +1090,23 @@ def grad_minus_reduced_likelihood_function(log10t):
...
                         constraints.append(
                             lambda log10t: noise_bounds[1]
-                            - log10t[i + len(self.theta0)]
+                            - log10t[v2]
                         )
                         bounds_hyp.append(noise_bounds)
-                try:
+                theta_limits = np.repeat(np.log10([theta_bounds]),repeats=len(theta0), axis=0)
+                sampling = LHS(xlimits=theta_limits,criterion="maximin",random_state=41)
+                theta_all_loops = sampling(self.options["n_multistart"])
```
relf, line 1107: I would suggest to np.vstack the theta0 from the user and the theta0s from the multistart (LHS) and just loop over the result. It should allow to suppress the "minimize duplication" and the optimal_theta_res_2?

relf (changes requested):
Thanks for the contribution, it definitely improves the results of kriging based models. 
I suggest some changes before merging (you should also pass black on the modified files)

smt/surrogate_models/krg_based.py
```diff
@@ -1082,18 +1090,23 @@ def grad_minus_reduced_likelihood_function(log10t):
                     )
 
                     for i in range(len(self.noise0)):
+                        v= i + len(self.theta0)
+                        v2=deepcopy(v)
```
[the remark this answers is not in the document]
Paul-Saves: v2 is an pointer of int but these pointers arise problems with multistart when calling the constraint.

I have not investigating the source of the problem since the deepcopy was a good go-around.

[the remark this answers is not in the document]
Paul-Saves: Ok, I will see that, I've problems with the new implementation

smt/surrogate_models/krg_based.py
```diff
@@ -1062,6 +1062,10 @@ def grad_minus_reduced_likelihood_function(log10t):
             self.D = self._componentwise_distance(D, opt=ii)
 
             # Initialization
+            try:
+                del optimal_theta_res
```
relf, line 1066: I think that deleting a variable is definitely a code smell. We certainly need to refactor the whole thing to avoid keeping an undesired state.

smt/surrogate_models/krg_based.py
```diff
@@ -1082,59 +1094,69 @@ def grad_minus_reduced_likelihood_function(log10t):
                     )
 
                     for i in range(len(self.noise0)):
+                        v = i + len(self.theta0)
+                        v2 = deepcopy(v)
```
relf, line 1098: Sorry, I still do not get that one. `deepcopy` should not be needed because in Python interpreter:
```python
>>> a = 1
>>> b = a
>>> a = 2
>>> b
1
>>> a
2
```
Moreover `v` is just used in that deepcopy?!

smt/surrogate_models/krg_based.py
```diff
@@ -1082,59 +1094,69 @@ def grad_minus_reduced_likelihood_function(log10t):
...
+                                    "maxiter": limit,
+                                },
+                            )
+                            try:
+                                if (
+                                    optimal_theta_res_loop["fun"]
+                                    < optimal_theta_res["fun"]
+                                ):
+                                    optimal_theta_res = optimal_theta_res_loop
+                            except:
```
relf, line 1133: I suppose you test for an exception because `optimal_theta_res` is not defined? For me it would be better to initialize it with something like:
```python
optimal_theta_res = { "fun": float("inf") }
```
But I do not know if it fits with the rest of the algorithm

relf (changes requested):
This is not specifically related to this PR:  `_optimize_hyperparams` is too big and we certainly need to extract some parts of it into smaller, more manageable methods. 
Tell me whether you can address the comments otherwise I will try to work on the refactoring.

smt/surrogate_models/krg_based.py
Paul-Saves: To be honest, I do not know why I have such a problem, what you say seems coherent. 
I will see if I could find another way to make it work.

Paul-Saves:
> This is not specifically related to this PR: `_optimize_hyperparams` is too big and we certainly need to extract some parts of it into smaller, more manageable methods.
> Tell me whether you can address the comments otherwise I will try to work on the refactoring.

I agree that the current code is not be the best, to say the least.

smt/surrogate_models/krg_based.py
Paul-Saves: Yes, I agree, the variables should not be keep in memory. It causes problem for KPLSK because we do two optimization cake some "trash" from the first interact with the second one.

smt/surrogate_models/krg_based.py
Paul-Saves: I thought it would be worse. I will do this, then. We definitely need a cleaner code. 
It was a pain to debug. There is 2 intricated loop for kriging and 4  for MFKPLSK...

smt/surrogate_models/krg_based.py
Paul-Saves: The problem dissapeared. This issue is then closed.

relf (approved):
Much better! Next PR related to `kriging_based.py` file should contain a refactoring of `_optimize_hyperparam`! :sweat_smile:

PR #296, opened 2021-03-02
adding random_state to the nestedLHS

anfelopera:


relf:
Thanks, I forgot that one. It seems like the MFK, MFKPLS and MFKPLSK tests that use NestedLHS are not too sensitive to randomness. Would you make them deterministic in this PR using your addition?

relf:
Looks good, do you plan to add something else?

anfelopera:
> Looks good, do you plan to add something else?

Nope, it's everything alright! ;)

PR #299, opened 2021-03-27
adapting gower distance for EGO and writing documentation

Paul-Saves:


relf (changes requested):
Thanks for adding this. A test with the option enabled shoud be added before merging.

PR #301, opened 2021-04-12
allowing n_start = 1

anfelopera:


relf (changes requested):
Thanks. Would you mind adding a simple test with n_start=1 just to prevent further regression?

relf (approved):

PR #317, opened 2021-06-09
Correct documentation typographic erros and apply black on files

Paul-Saves:


Paul-Saves:
KPLS is still false !

doc/_src_docs/kpls.rstx
```diff
@@ -0,0 +1,27 @@
...
+
+Usage
+-----
+
+.. embed-test-print-plot :: smt.surrogate_models.tests.test_surrogate_model_examples , Test , test_kpls , 80
+
+Options
+-------
+
+.. embed-options-table :: smt.surrogate_models , KPLS , options
```
relf, line 27: I suppose you have not changed the whole file... Do you know why it appears has newly added?  (same thing for krg.rstx)

Paul-Saves: It is because the file is in the wrong folder, sorry

Paul-Saves: Thank you for the proof-reading :) 
The files were misplaced at the root of the documentation

relf (approved):

PR #323, opened 2021-07-09
Dev rconde - Expanded LHS, new "ese" method

rconde1997:
Good morning,
I made some changes in the "expanded LHS" function . Basically, the objective was to implement the "ese" optimization method to expand a Latin Hypercube Sampling (LHS). To do that, I saw that there was already a variable, so-called "fixed_index" that allow us to optimize a LHS but by fixing some of the points (what I wanted). The problem was that the variable was there but it was unused. 

As a consequence, I made a few changes in the __ese_ function to use this variable. And finally I added the "ese" optimization to the function _expand_lhs_ .

In order to get a better understanding of the process, I have uploaded a tutorial to show a brief example on how to use the functionality.


smt/sampling_methods/tests/test_lhs.py
```diff
@@ -44,7 +44,7 @@ def test_expand_lhs(self):
             new = int(new_list[i])
             new_num = num + new
 
-            x_new = sampling.expand_lhs(x, new)
+            x_new = sampling.expand_lhs(x, new, method="ese")
```
relf, line 47: Assertions below did not change. Could you add something related to the fact that "ese" is used?

relf (changes requested):

Issue #313, opened 2021-05-11
Limitation with mixed integer variables (can not treat all types of variables)


Paul-Saves:
Hello, 
If I have an integer variable, that is quantitative, like X ∈ { 1,2,3}, it works but for a variable that is of the form  X ∈ { 1, 2.18 ,4} or  X ∈ { 1,2.5}, there is no solution yet.

The workaround are; creating a wrapper or treat these variables as categorical but these solution are bad because they are either not computing properly the distance or more expensive. 

The enhancement would be to replace INT type for integer variables into a more general ORD type for ordinal values

Paul-Saves:
Still in limitations, with categorical outputs, only constant regression can be used.

Paul-Saves:
Solved by #326

PR #326, opened 2021-07-30
INT type in mixed integer models is now about ordered variables 

Paul-Saves:
Hello, 
If I have an integer variable, that is quantitative, like X ∈ { 1,2,3}, it works but for a variable that is of the form  X ∈ { 1, 2.18 ,4} or  X ∈ { 1,2.5}, there is no solution yet.

The workaround are; creating a wrapper or treat these variables as categorical but these solution are bad because they are either not computing properly the distance or more expensive. 

This PR update INT type for integer variables into a more general type for ordinal values


Paul-Saves:
fix #313

smt/applications/mixed_integer.py
```diff
@@ -24,11 +25,22 @@ def check_xspec_consistency(xtypes, xlimits):
 
     for i, xtyp in enumerate(xtypes):
         if (not isinstance(xtyp, tuple)) and len(xlimits[i]) != 2:
-            raise ValueError(
-                "Bad x limits ({}) for variable type {} (index={})".format(
-                    xlimits[i], xtyp, i
+            if xtyp == "int_type" and isinstance(xlimits[i][0], str):
```
relf, line 28: `if xtyp == INT and`...

smt/applications/mixed_integer.py
```diff
@@ -24,11 +25,22 @@ def check_xspec_consistency(xtypes, xlimits):
...
+                listint = list(map(float, xlimits[i]))
+                sortedlistint = sorted(listint)
+                if np.array_equal(sortedlistint, listint) == False:
```
relf, line 31: `if not np.array_equal(sortedlistint, listint):` ...

smt/applications/tests/test_mixed_integer.py
```diff
@@ -107,11 +146,106 @@ def test_cast_to_mixed_integer(self):
...
+        l = unfold_xlimits_with_continuous_limits(xtypes, xlimits)
+        self.assertEqual(
+            np.array_equal(
+                [[-5, 5], [0, 1], [0, 1], [0, 1], [0, 1], [0, 1], [0, 2]],
+                unfold_xlimits_with_continuous_limits(xtypes, xlimits),
+            ),
+            True,
+        )
+
+    def test_unfold_xlimits_with_continuous_limits2(self):
```
relf, line 169: Better to suffix the test name with some information about the specificity of the test, like `test_unfold_xlimits_with_continuous_limits_and_ordinal_values`

relf (changes requested):
We should add a new type (like `ORD`) to represent the new ordered variable capability (and keep INT for integer variables), as it is a bit misleading to have INT type for {"0", "3.5"} values.

Paul-Saves:
Can't we just keep ORD and deprecate old INT type as ORD generalize INT ?

relf:
> 
> 
> Can't we just keep ORD and deprecate old INT type as ORD generalize INT ?

Yes sure! Deprecation means it is still working in at least in a next minor release (SMT 1.1) and ideally is removed in next major release (well I do not think we aim at a SMT 2.0 just for that).

relf:
Well, even if ORD can supersede INT, would not it be interesting to keep INT anyway (no deprecation then) to have clearer/cleaner specification ([1, 2, 3] vs ["1", "2", "3"]) when we have integer variable?

Paul-Saves:
I think INT should be removed. It makes no sense to have both INT and ORD as the variables are treated similarly.

relf:
Ok then could you just keep a test with INT to ensure it still works (we will remove all this later on after SMT 1.1).

relf:
> I think INT should be removed later. It makes no sense to have both INT and ORD as the variables are treated similarly.

It was just a matter of user-facing API, but ok it is not worth the code complexity.

relf:
Looks good to me, anything to add?

Paul-Saves:
> 
> 
> Looks good to me, anything to add?

I don't think so, it's ok for me

PR #333, opened 2021-09-01
fix bug in import

Paul-Saves:
Fix a circular import in smt tutorial


smt/surrogate_models/krg_based.py
```diff
@@ -26,8 +26,10 @@
 )
 from scipy.stats import multivariate_normal as m_norm
 from smt.sampling_methods import LHS
-from smt.applications.mixed_integer import GOWER
-
+try : 
+    from smt.applications.mixed_integer import GOWER
```
relf, line 30: I did not catch this before, but `surrogate_models` must not depend on `applications`! `applications` depends on `surrogate_models` hence `GOWER` must be defined here.

relf (changes requested):

PR #335, opened 2021-11-16
update pls auto and add it to notebook

Paul-Saves:


smt/surrogate_models/kpls.py
```diff
@@ -102,31 +104,33 @@ def _estimate_number_of_components(self):
                 # as the problem is the same and the data are mainly similar
                 if fold == 0:
                     super(KPLS, self)._initialize()
+                    self.options["corr"] = corr
                 else:
                     super(KPLS, self)._initialize()
+                    self.options["corr"] = corr
```
relf, line 110: Looking rapidly, not sure to understand why `self.options["corr"]` is assigned but the assignment could be factorized out after the if and the `super(KPLS, self)._initialize()` factorized out before.

smt/surrogate_models/kpls.py
```diff
@@ -102,31 +104,33 @@ def _estimate_number_of_components(self):
...
                 nextcomp = False
-
+        ncomp = self.options["n_comp"]
         self.training_points[None][0][0] = X
         self.training_points[None][0][1] = y
         self.nt = len(X)
         super(KPLS, self)._initialize()
+        self.options["corr"] = corr
+        self.options["n_comp"] = ncomp
+        print("\n The new number of PLS component is", ncomp)
```
relf, line 133: Maybe component(s) and add a space after 'is'

relf (changes requested):

smt/surrogate_models/kpls.py
```diff
@@ -100,33 +102,32 @@ def _estimate_number_of_components(self):
                 self.training_points[None][0][1] = yfold
                 # We are doing hyperparameters hot_start for the k-fold
                 # as the problem is the same and the data are mainly similar
-                if fold == 0:
-                    super(KPLS, self)._initialize()
-                else:
-                    super(KPLS, self)._initialize()
+                super(KPLS, self)._initialize()
+                self.options["corr"] = corr
```
relf, line 106: Hi Paul. Could you elaborate on this change? I understand that you've saved `corr` option and re-assign it here because the `_initialize` call re-declares options and resets it to the default value but in that case we should do the same thing with all options? Or did I miss something? Finally why do you call `_initialize` ? Maybe you should reset relevant fields only?

PR #340, opened 2021-12-22
Add Gradient Enhanced EGO optimization

Laurentww:
New feature enabling (indirect) Gradient Enhanced EGO optimization by combining the existing GEKPLS and EGO classes.


relf:
Hi, thanks for your contribution. You will have to rebase your PR as master has new commits.
Your test passes with 3.6 but fails with 3.7, we have to check differences in the used dependencies.
Python 3.6 has just reached its [end of life date](https://endoflife.date/python), I noticed that numpy >= 1.20 does not support 3.6. Does it explain the failure? In your environment, do/could you test against recent version numpy 1.21.5 and scipy 1.7.3?

Laurentww:
Thanks for checking. I updated the ego_gek test assertion, the result values were flipped. The tests pass with numpy 1.21.5 and scipy 1.7.3.

smt/applications/tests/test_ego.py
```diff
@@ -464,6 +464,45 @@ def test_find_best_point(self):
...
+                assert kx is None
+                _self = super(TensorProductIndirect, self)
+                response = _self._evaluate(x, kx)
+                sens = np.hstack(_self._evaluate(x, ki) for ki in range(x.shape[1]))
+                return np.hstack((response, sens))
+
+        fun = TensorProductIndirect(ndim=2, func="tanh")
+
+        # Construction of the DOE
+        sampling = LHS(xlimits=fun.xlimits, criterion="m")
```
relf, line 481: Maybe you should set `random_state` parameter when creating LHS instance as well to get result reproducibility.
```python
sampling = LHS(xlimits=fun.xlimits, criterion="m", random_state=0)
```
then fix the test assertion if needed

Laurentww:
I kept the `random_state` unchanged because sufficiently converged design variables should not depend on the DOE. 

The test optimization function is now exponential. For which the design variables should always converge to -1.

relf:
Looks good to me. You have to update your branch with changes from master and add your name as a contributor in the `AUTHORS.md` file.

Laurentww:
Thanks for your effort into this PR. I have merged the recent commits into this branch and updated the author list with my name.

PR #346, opened 2022-01-25
Fix EI to work with 2D `y_data` as used in GEKPLS

Laurentww:
Fix for  PR #340. `EI()` Now correctly identifies the optimal design vector from `y_data`

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -15,6 +15,7 @@ SMT has been developed thanks to contributions from:
 * Frederick Zahle
 * Jasper Bussemaker
 * Julien Schueller
+* Laurent Wilkens
 * Lucas Alber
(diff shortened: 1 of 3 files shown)


relf:
Thanks for the fix. Would you add a test with EI to check it works (regression test for further developments)?

Laurentww:
Added the test. 

Also made the following two other code changes, which simplify testing of EI. Let me know if these need a separate PR or are not desired.
 - Removed the `y_data` argument `ego.EI()` . `y_data` now is read from the gpr, which contains the same `y-data`
 - Refactored training of gpr into private ego method

Laurentww:
Updated the PR. Thanks for suggesting the assertion fix!

PR #367, opened 2022-06-08
change warning message from print to warnings.warn()

zhoutianxun:
Currently for KRG models, warning messages are printed. The user is unable to suppress the warnings using a warning filter.

**Changes**
The three printed warning messages are changed to warnings.warn(). This allows the user to suppress the warning message using a warning filter if necessary.
Warning is formatted to print only the message, same as before.

### DIFF
--- a/smt/surrogate_models/krg_based.py
+++ b/smt/surrogate_models/krg_based.py
@@ -6,6 +6,7 @@ This package is distributed under New BSD license.
 import numpy as np
 from scipy import linalg, optimize
 from copy import deepcopy
+import warnings
 
 from smt.surrogate_models.surrogate_model import SurrogateModel
 from smt.utils.kriging_utils import differences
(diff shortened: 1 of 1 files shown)


Paul-Saves:
There is a typo on l. 1995
`arnings.warn("Warning: noise0 is out the feasible bounds. The lowest possible value is used instead.\n")`

smt/surrogate_models/krg_based.py
```diff
@@ -1087,9 +1090,10 @@ def grad_minus_reduced_likelihood_function(log10t):
                             or self.noise0[i] > noise_bounds[1]
                         ):
                             self.noise0[i] = noise_bounds[0]
-                            print(
-                                "Warning: noise0 is out the feasible bounds. The lowest possible value is used instead."
-                            )
+                            default_warn = warnings.formatwarning
+                            warnings.formatwarning  = lambda msg, *args, **kwargs: str(msg)
+                            arnings.warn("Warning: noise0 is out the feasible bounds. The lowest possible value is used instead.\n")
```
relf, line 1095: As @Paul-Saves mentioned, there is a typo here

relf (changes requested):
Thanks for your contribution. 

I cannot think of an actual reason to keep the format, so I suggest to simply use: `warnings.warn()` instead of `print()` (without `formatwarning` trick)

smt/surrogate_models/krg_based.py
zhoutianxun: Thank you for spotting! I will update to just use warnings.warn() without the formatting

relf (approved):

PR #369, opened 2022-06-08
Airfoil Parameters surrogate model example

raul-rufato:
I am implementing an example of a surrogate model that uses SMT to be able to determine the aerodinqmic parameters of a given aerofoil, such as cd,cl,cm. And it uses the mode shapes, Mach number, angle of attack and reynolds number as inputs. I believe the best place to put this example is in this section, if not please tell me. The example was implemented in colab and notebook and all the necessary files are attached.


relf:
Hi Raul. Thanks for this contribution however I would prefer to have your example as a Python script like the other ones. It allows us to include them in the documentation [here](https://smt.readthedocs.io/en/latest/_src_docs/examples.html#examples).
Moreover examples are executed when documentation is generated checking automatically they are not broken.

relf:
Hi Raul. Thanks for this contribution however I would prefer to have your example as a Python script like the other ones. It allows us to include them in the documentation [here](https://smt.readthedocs.io/en/latest/_src_docs/examples.html#examples).
Moreover examples are executed when documentation is generated checking automatically they are not broken.

PR #389, opened 2022-12-10
add output derivative for every kernel and fix a major issue for linear and quadratic trends

Paul-Saves:
- Add predict output derivatives for abs_exp, matern32 & matern52.
- Add tests for the new kernels derivatives.

- Fix a major issue for linear or quadratic trends (see below).
-  Correct a bug in matplotlib 5  in the gekpls documentation test.
- Add a brand new test to avoid breaking the trends/poly in the future.


Paul-Saves:
Kriging with linear "poly" BEFORE the patch (issue) : 
image

Kriging with linear "poly" AFTER the patch (corrected) : 
image

smt/surrogate_models/tests/test_krg.py
```diff
@@ -88,14 +103,16 @@ def _check_derivatives(sm, xt, yt, ndim, i=10):
...
-            dydx_predict[j] = sm.predict_derivatives(x_test, kx=j)[0]
-            dydx_predict[j] = sm.predict_derivatives(x_test, kx=j)[0]
-            dydx_predict[j] = sm.predict_derivatives(x_test, kx=j)[0]
+
         print(dydx_predict)
         print(yt[i, 1:])
 
         # compare results
-        np.testing.assert_allclose(yt[i, 1:], dydx_predict, atol=2e-3, rtol=1e-3)
+        if sm.options["corr"] == "squar_exp":
```
relf, line 111: Seen with Paul, weird test to be suppressed. Seems we are comparing true yt values and its derivatives while test function being Sphere?!

Paul-Saves: fixed

PR #394, opened 2023-01-02
Correct Gower distance bouds inconsistency & EGO for mixed integer

Paul-Saves:
- There was both an inconsistency for the data normalization within Gower distance and a problem as the data were not in [0,1] as intended.  
This PR correct this error by using directly the bounds of the variables (xlimits) and by doing a proper job to project the data into [0,1]^n. 

- The  Acquisition criterion was not maximized at all with Mixed-integer variables. Now, it is maximized from random starting points by CObyLA. 

Fix several bugs.

### DIFF
--- a/smt/applications/ego.py
+++ b/smt/applications/ego.py
@@ -374,19 +374,19 @@ class EGO(SurrogateBasedApplication):
             cons = []
             for j in range(len(bounds)):
                 lower, upper = bounds[j]
-                if isinstance(self.options["xtypes"][j], tuple):
-                    upper = int(upper - 1)
+                if self.options["categorical_kernel"] is not None:
+                    if isinstance(self.options["xtypes"][j], tuple):
+                        upper = int(upper - 1)
                 l = {"type": "ineq", "fun": lambda x, lb=lower, i=j: x[i] - lb}
                 u = {"type": "ineq", "fun": lambda x, ub=upper, i=j: ub - x[i]}
                 cons.append(l)
                 cons.append(u)
-                options = {"tol": 1e-6,"rhobeg":0.1}
+                options = {"catol": 1e-6, "tol": 1e-6, "rhobeg": 0.1}
         else:
             bounds = self.xlimits
             method = "SLSQP"
             cons = ()
             options = {"maxiter": 200}
-
 
         if criterion == "EI":
             self.obj_k = lambda x: -self.EI(np.atleast_2d(x), enable_tunneling, x_data)
@@ -436,6 +436,7 @@ class EGO(SurrogateBasedApplication):
         x_et_k = np.atleast_2d(opt["x"])
 
         return x_et_k, True
+
     def _get_virtual_point(self, x, y_data):
         """
         Depending on the qEI attribute return a predicted value at given point x
--- a/smt/applications/ego.py
+++ b/smt/applications/ego.py
@@ -428,7 +428,6 @@ class EGO(SurrogateBasedApplication):
             if not success:
                 self.log("New start point for the internal optimization")
                 n_optim += 1
-                n_start += 10
         if n_optim >= n_max_optim:
             # self.log("Internal optimization failed at EGO iter = {}".format(k))
             return np.atleast_2d(0), False
--- a/smt/applications/ego.py
+++ b/smt/applications/ego.py
@@ -381,8 +381,8 @@ class EGO(SurrogateBasedApplication):
                 u = {"type": "ineq", "fun": lambda x, ub=upper, i=j: ub - x[i]}
                 cons.append(l)
                 cons.append(u)
-                options = {"catol": 1e-6, "tol": 1e-6, "rhobeg": 0.1}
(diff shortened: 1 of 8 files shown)


Paul-Saves:
Hi, @relf ,
 I have found several bugs recently while working on hierarchical variables Kriging models.
This PR fixes the latter problems. ( I can't ask for review prior to 1st review. ) 

Paul

smt/applications/ego.py
```diff
@@ -403,7 +428,7 @@ def _find_best_point(self, x_data=None, y_data=None, enable_tunneling=False):
             if not success:
                 self.log("New start point for the internal optimization")
                 n_optim += 1
-
+                n_start += 10
```
relf, line 431: Why do you increase the `n_start` ? I mean, as `n_start` and even `n_max_optim` are controlled by the user to achieve give optimization more chances, is it required?

smt/applications/ego.py
```diff
@@ -298,13 +298,22 @@ def _setup_optimizer(self, fun):
                 random_state=self.options["random_state"],
                 output_in_folded_space=work_in_folded_space,
             )
+            self._sampling_optim = self.mixint.build_sampling_method(
```
relf, line 301: Is `self._sampling_optim` actually needed when we already have `self._sampling` ? We might also need to set the `random_state` to get reproducibility.

relf (commented):
Thanks Paul! I have a few questions...

smt/applications/ego.py
Paul-Saves: - We want to have truly random points to avoid being stuck in the same 20 neighbourhoods again and again. 

- We want to have truly random points so that when we are stuck at an iteration, we are not stuck at the next one.

That is why we do not want seeded starting points for optimization.  

On the other hand, the initial DoE is generating only once, so it makes sense to have a seed for it.

relf: Yeah, actually I was guessing such explanation. This is a bug of the current implementation when random_state is set.

The fact is that the current implementation of LHS yields the same sampling again and again when random_state is set.

A better behaviour for sampling method would be to generate different samplings when random_state is set, but it would yield the same sequence of samplings. I will open an issue for this. 

Actually, if we increase `n_start` like you did, it is a way to workaround the problem without introducing `self.sampling_optim` and loosing the reproducibility. What do you think?

Paul-Saves: Yes, I got your point. 
Increasing `n_start` was not a good solution, to be honest. The problem is that we want different starting points at every iteration. 
If we want to have 50 points added by EGO, incresing `n_start` 50 times do not seems reasonable. Even more if we we also increase this number everytime the optimization failed. 
We would just end up with an optimization process that takes hours...

relf: Ok I got it. Let's merge!

smt/surrogate_models/ls.py
```diff
@@ -28,7 +28,11 @@ def _initialize(self):
         super(LS, self)._initialize()
         declare = self.options.declare
         supports = self.supports
-
+        declare(
```
relf, line 31: Forgot this one! Why did you add `xlimits` option to `LS` and `QP`? It does not seem to be used.

Paul-Saves: In Mixed integer surrogate model, we have: 
        self._surrogate.options["xlimits"] = self._xlimits
Therefore, I put xlimits in every surrogate.

The question is: Why aren't the limits used to normalized data in LS & QP ? I haven't investigate LS or QP but I assume the data should also be normalized according to xlimits.

relf: Well... You add this line and I realize now that you introduce `xlimits` in `krg_based` to fix gower distance (I was fooled by this familiar name which is used by sampling methods). But so far surrogate models did not need `xlimits`! 

Moreover why just LS and QP and not the others IDW, RBF, etc. If we add it to all it has to be done in `SurrogateModel` class. But I do not think this is the right thing to do here. 

Gower distance is only used with kriging based models, and `xlimits` is only used with Gower distance. So for me it is a very specific option and should be named something like `gower_xlimits` to make it clear and `MixedIntegerSurrogateModel` should not set it (the surrogate should have been initialized properly with the right `gower_xlimits` when given to `MixedIntegerSurrogateModel`). Could you open a new PR for this ?

Paul-Saves: Yes, I'll definitively check that out and open the PR when it's ready ! 

- I agree and I removed xlimits in QP & LS and have clean some stuff. 

- The mixed integer parameter "`xlimits`" is the same that is used for the sampling. It is also the same that we use for optimization and for the decorator.
- It makes sense to me that it keep its name. Typically, for optimization, the API do not ask to write `xlimits `in the parameters of both the optimizer, the decorator and the model. We could change all that for xtypes also but it would be another topic, I think.

- I also realized that the data were not normalized for QP (LS seems fine).

PR #396, opened 2023-01-11
Add data normalization in QP and clean QP, LS and EGO

Paul-Saves:
Add data normalization in QP
xlimits is no longer a parameter of QP and LS
self.work_in_folded_space is now generalized and properly checked

### DIFF
--- a/smt/applications/mixed_integer.py
+++ b/smt/applications/mixed_integer.py
@@ -104,7 +104,8 @@ class MixedIntegerSurrogateModel(SurrogateModel):
         self._cat_kernel_comps = cat_kernel_comps
         self._xtypes = xtypes
         self._xlimits = xlimits
-        self._surrogate.options["xlimits"] = self._xlimits
+        if "xlimits" in self._surrogate.options:
+            self._surrogate.options["xlimits"] = self._xlimits
 
         self._input_in_folded_space = input_in_folded_space
         self.supports = self._surrogate.supports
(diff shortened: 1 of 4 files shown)


Paul-Saves:
Thanks to @relf, following issues of #394

smt/applications/mixed_integer.py
```diff
@@ -104,7 +104,8 @@ def __init__(
         self._cat_kernel_comps = cat_kernel_comps
         self._xtypes = xtypes
         self._xlimits = xlimits
-        self._surrogate.options["xlimits"] = self._xlimits
+        if "xlimits" in self._surrogate.options:
+            self._surrogate.options["xlimits"] = self._xlimits
```
relf, line 108: I am still puzzled by this `xlimits` option. Why not use the values themselves to normalize when using Gower distance? Like `standardization`?

smt/surrogate_models/qp.py
```diff
@@ -11,7 +11,7 @@
 import scipy
 from smt.surrogate_models.surrogate_model import SurrogateModel
 from smt.utils.caching import cached_operation
-
+from smt.utils.kriging_utils import standardization
```
relf, line 14: Well, `standardization` should be moved in a another module in utils? `utils/misc.py` ?

smt/surrogate_models/qp.py
```diff
@@ -118,7 +118,7 @@ def _predict_derivatives(self, x, kx):
             Derivative values.
         """
         dim = self.nx
-
+        x = (x -self.X_offset)/self.X_scale
```
relf, line 121: Seems you need to pass `black`.

relf (changes requested):

Paul-Saves:
Ok, j'ai passé `black`.  Le reste sera traité dans la prochaine PR

PR #411, opened 2023-03-01
Exponential correlation function with variable power

yqliaohk:
I added an option to use exponential correlation function with a variable power.
I kept the original `squar_exp` and `abs_exp` to maintain potential backward compatibility, but I think these three can be combined and re-organized to reduce duplication if this won't introduce backward incompatibility or it is not a concern.
I can work on combining them if the smt team agrees this is ideal.
The tests passed locally. Exceptions are using this exponential function with a power other than 2 for `test_kpls_auto.py`. But I also noticed it failed when using original `abs_exp`. If this is expected, I can leave as it is. If not, I can try to fix it maybe with someone on team to help me better understand this test.
Please let me know any suggestions on additional tests.


relf:
Hi Yingqian, thanks for your contribution. Indeed it is a generalization and it would be nice to refactor and remove code duplication. 

Nevertheless we have to assess the impact of changing the string enumeration (if we go down this line we could also use a Python enum). I am always a bit reluctant to loose backward compatibility, but at the same time we are preparing a 2.0, so it is definitely the right time to break things! 😅

relf (changes requested):
Ok, I suggest to keep backward compatibility while using your generalized implementation and remove duplication.

* First I would rename the new value `exp` in `pow_exp` (as there is also an `act_exp`) and the option `power` in `pow_exp_power` to be more specific.
* On initialization, if `abs_exp` or `squar_exp` is set, then automatically set `corr` option to `pow_exp` and option `pow_exp_power` with relevant power value.
* You noticed that some conditionals are done against `squar_exp` or `abs_exp` meaning they have to be replaced by the equivalent tests on `pow_exp_power`.

yqliaohk:
Hi Rémi, thank you for your review! These are great suggestions. I will make corresponding changes. (Expected time: by Mar 5)

yqliaohk:
I have tried refactoring them based on the suggestions. I had some difficulties to implement exactly the same for the second suggestion because the option can still be changed after initialization, so I think I still need to keep the value update part after initialization. Let me know what you think and if you have other suggestions!

smt/surrogate_models/kplsk.py
```diff
@@ -23,13 +23,25 @@ def _initialize(self):
...
 
+        declare(
+            "pow_exp_power",
+            2.0,
+            desc="Power for the pow_exp kernel function",
+            types=(float),
+        )
+
+
     def _componentwise_distance(self, dx, opt=0, theta=None, return_derivative=False):
```
Paul-Saves, line 34: The value pow_exp_power should be > 0 and =<2.

The kernel function $\phi(x) = \exp [- \theta |x|^p]$ is positive definite for a given positive $\theta$ if $0<p \leq 2$ [Corollary~3]{Schoenberg}. METRIC SPACES AND POSITIVE DEFINITE FUNCTIONS* BY I. J. SCHOENBERG

Paul-Saves (changes requested):

smt/surrogate_models/kplsk.py
yqliaohk: Thank you for pointing it out. I added a check. Please let me know if any additional change is needed.

smt/surrogate_models/krg_based.py
```diff
@@ -78,6 +81,12 @@ def _initialize(self):
             ),
             desc="Correlation function type",
         )
+        declare(
+            "pow_exp_power",
+            3.0,
```
Paul-Saves, line 86: Can you initialize to a value other than 3 ? Let's say 1.5 ?

smt/surrogate_models/kplsk.py
Paul-Saves: Thanks a lot !

smt/surrogate_models/krg_based.py
yqliaohk: Yes! sorry I tested different default values locally but forgot to change it back.

Paul-Saves (approved):

smt/surrogate_models/krg_based.py
```diff
@@ -78,6 +81,12 @@ def _initialize(self):
             ),
             desc="Correlation function type",
         )
+        declare(
+            "pow_exp_power",
+            1.9,
+            types=(float),
+            desc="Power for the pow_exp kernel function, values (0.0, 2.0], for squar, abs, and matern, this will be initialized correspondingly by default",
```
relf, line 88:
```suggestion
            desc="Power for the pow_exp kernel function (valid values in ]0.0, 2.0]). This option is set automatically when corr option is squar, abs or matern.",
```

relf (changes requested):

smt/applications/mfk.py
```diff
@@ -130,6 +130,12 @@ def _initialize(self):
         )
         self.name = "MFK"
 
+        # initialize default power values
+        if self.options["corr"] == "squar_exp":
```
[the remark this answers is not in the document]
yqliaohk: Good catch! sorry I forgot about this one.

smt/surrogate_models/kpls.py
```diff
@@ -25,7 +25,7 @@ def _initialize(self):
         declare(
             "corr",
             "squar_exp",
-            values=("abs_exp", "squar_exp"),
+            values=("abs_exp", "squar_exp"), # Should add pow_exp?
```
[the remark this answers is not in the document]
yqliaohk: I add tests in `test_kpls.py`, let me know if you think that isn't enough.

relf (approved):

relf:
@yqliaohk Thanks for your work and this nice contribution!

yqliaohk:
> @yqliaohk Thanks for your work and this nice contribution!

My pleasure to contribute! Thank you both for the comments and suggestions!

PR #416, opened 2023-03-21
fix Goldstein test & add cantilever and neural network test into SMT/Problems.

Paul-Saves:


smt/problems/neural_network.py
```diff
@@ -0,0 +1,62 @@
...
+Multi-Layer Perceptron problem from:
+C. Audet, E. Hall e-Hannan, and S. Le Digabel. A general mathematical framework for constrained mixed-variable blackbox optimization problems with meta and categorical variables. Operations Research Forum,499
+4:137, 2023.
+    """
+import numpy as np
+
+from smt.problems.problem import Problem
+
+
+class HierNN(Problem):
```
relf, line 14: As it is a benchmarking `Problem`, may be we can go with the full name : `HierarchicalNeuralNetwork` for the class name

PR #428, opened 2023-04-25
New design space API

jbussemaker:
Implementation of a new design space definition API in `smt.utils.design_space`:

- One API for all (mixed-discrete, hierarachical) design space definition needs: variables, hierarchical structure, value constraints; all in correct OOP fashion
- `BaseDesignSpace` class implementing interfaces for:
  - Sampling, design vector correction, requesting the acting-status of design variables (i.e. whether they are acting or non-acting, relevant for hierarchical Kriging), folding/unfolding/rounding
  - This class can be extended if any user of smt has their own implementation for the sampling and/or correction logic
- `DesignSpace` class implementing the new API that should be pushed, using [ConfigSpace](https://automl.github.io/ConfigSpace/) as its underlying design space model for implementing hierarchy and value constraints
  - ConfigSpace becomes a new dependency of smt (it's a very active and very well maintained library!)
- `LegacyDesignSpace` class and `ensure_design_space` function for converting any old format x definition (xlimits, xtypes, xspec) into a design space
- Added tests

This implementation using ConfigSpace as the underlying hierarchical design space model means that the conditionality relationships are now a directed acyclic graph, most importantly enabling *nested conditionality*.

Integration of the new design space API in SMT:

- Integrated in krg_based
  - krg_based now takes `design_space` as option (`xspecs` is converted to a design space if given)
  - Mixed-discrete and hierarchical kernels are updated to use the new design space object
- Mixed-integer and EGO applications use the new design space API
- `Problem` class now has a design space definition (coming from `xlimits` if not explicitly set)
- Updated the documentation (coming in a separate pull request)

PS: I'm not entirely sure why these 2 commits from Paul are in this pull request... I must have done something wrong when updating my fork


jbussemaker:
Additions:
- Removed dependency on ConfigSpace for now
- DesignSpace can be instantiated with continuous xlimits
- XSpecs and XType have been completely replaced by DesignSpace
- Added numba speedup: numba is an optional dependency and only affects Kriging calculations, nothing the user should ever interface with
- A small bug fix in KPLS

relf (commented):
Thanks Jasper. In a first pass, I am questionning the added argument `is_acting=None` to surrogate_models API methods:

* Do we imagine/plan to develop hierarchical variables handling for other surrogates than kriging-based ones? 
* Another way (less functional more stateful) would be to add a `set_acting_mask()` like we have a `set_training_values()`? Depending on how the API is used such way could be awkward but it is less intrusive for existing method signatures. 
* If we go for all surrogates, in `SurrogateModel` we have the `supports` dictionary which allows to declare supported features, it is was till now derivatives oriented but we could add something like : supports["hierarchical_variables"], supports["categorical_variables"]

@NatOnera, @PaulSaves, any thoughts on this?

jbussemaker:
Very good point. The idea behind passing the `is_acting` matrix to the surrogate model when setting training values (which also works if we had a different function for this) is:
- In principle, the surrogate model also has the DesignSpace instance, so it could request this `is_acting` matrix also from there
- However, for complicated design space models this could incurr some extra overhead, so if this information is already available it is better to pass it to the surrogate model to prevent this overhead

I'd definitely be in favor of adding hierarchy to the `supports` dictionary.

jbussemaker:
`prange` bug is fixed but the commit somehow doesn't show up...

relf:
I've pulled the PR, test and got the `prange` error. Maybe you've pushed the bug fix on the wrong branch?

jbussemaker:
Ok I made a new commit now and the `prange` fix is also included

doc/_src_docs/applications/Mixed_Hier_usage_TestMixedInteger_run_mixed_integer_lhs_example.png
relf, line 1: Just noticing here, we do not have the same kind of result. To be checked...

doc/_src_docs/applications/ego_TestEGO_run_ego_mixed_integer_example.png
relf, line 1: We do not have the same convergence here. To be checked...

relf (approved):
Thank you @jbussemaker! This PR is a huge improvement of the handling of categorical and hierarchical variables. 

I've noticed some behaviour changes to be checked and a problem with an EGO test on windows (tried to re-run several times without success). It is definitely not a problem for me to merge regarding the code improvements brought in by this PR. @Paul-Saves, @NatOnera is it ok for you?  

@NatOnera, @Paul-Saves, notice also that `utils/mixed_integer.py` is removed here, something to be taken into account in downstream code SEGOMOE.

PR #431, opened 2023-05-16
 fix default options for categorical kernels  and  fix folding for ego

Paul-Saves:


Paul-Saves:
@relf @jbussemaker Good to me ! :)

PR #438, opened 2023-06-05
Fix test reproducibility

relf:
* Set random seed for initial DOE build. Note: does not work with 42.
* Add numba to requirements to benefit from JIT compilation


relf:
Seems that coverage decreases due to JIT usage, so ignore.

PR #441, opened 2023-06-05
Correcting bugs in SMT 2.0 

Paul-Saves:
Finished and okay for merging

### DIFF
--- a/smt/surrogate_models/krg_based.py
+++ b/smt/surrogate_models/krg_based.py
@@ -259,7 +259,7 @@ class KrgBased(SurrogateModel):
         y = self.training_points[None][0][1]
         # Get is_acting status from design space model if needed (might correct training points)
         is_acting = self.is_acting_points.get(None)
-        if is_acting is not None or self.design_space.is_all_cont == False:
+        if is_acting is not None or not (self.design_space.is_all_cont):
             X, is_acting = self.design_space.correct_get_acting(X)
             self.training_points[None][0][0] = X
             self.is_acting_points[None] = is_acting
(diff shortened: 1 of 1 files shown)


smt/surrogate_models/krg_based.py
```diff
@@ -1159,7 +1159,7 @@ def _predict_values(self, x: np.ndarray, is_acting=None) -> np.ndarray:
             Evaluation point output variable values
         """
         # Initialization
-        if is_acting is None:
+        if is_acting is not None or not (self.design_space.is_all_cont):
```
relf, line 1162: This is weird because you never use the `is_acting` array content given as argument.

smt/surrogate_models/krg_based.py
```diff
@@ -1319,7 +1319,7 @@ def _predict_variances(self, x: np.ndarray, is_acting=None) -> np.ndarray:
             Evaluation point output variable MSE
         """
         # Initialization
-        if is_acting is None:
+        if is_acting is not None or not (self.design_space.is_all_cont):
```
relf, line 1322: Same thing here.

relf:
I think you should add some tests (is_acting being None or something)  to reveal the bug then make them pass with the relevant fix.

smt/surrogate_models/krg_based.py
Paul-Saves: It is used by gower_componentwise_distances

relf: I mean you have changed the logic. Before your fix when `is_acting` was None  we retrieve it from ` self.design_space.correct_get_acting(x)`, otherwise we just use it.

Now if we pass an `is_acting` array, you seem to override it with `x, is_acting = self.design_space.correct_get_acting(x)`. So whether or not it is right, the intent is not clear and it looks a bit confusing.

Moreover as all the tests were successful before I think we should add some tests to cover the logic we want and document the expected behaviour as well.

PR #443, opened 2023-06-07
speeding up EGO

Paul-Saves:
- For EGO the default optimizer is SLSQP with 200 iterations at max for continuous variables.
Now, for EGO, the default optimizer is COBYLA with 200 iterations at max for mixed variables.

- For optimizing the Kriging likelihood, the maximal number of iterations is 10 times the number of hyperparameters with a minimum of 25 iterations.

- Now Numba is not used by default anymore but has become a facultative option that is given by the environment variable  USE_NUMBA_JIT.

### DIFF
--- a/smt/surrogate_models/krg_based.py
+++ b/smt/surrogate_models/krg_based.py
@@ -1513,7 +1513,7 @@ class KrgBased(SurrogateModel):
                 )
                 return res
 
-        limit, _rhobeg = 10 * len(self.options["theta0"]) + 5, 0.5
+        limit, _rhobeg = max(10 * len(self.options["theta0"]), 25), 0.5
         exit_function = False
         if "KPLSK" in self.name:
             n_iter = 1
--- a/smt/surrogate_models/krg_based.py
+++ b/smt/surrogate_models/krg_based.py
@@ -1513,7 +1513,7 @@ class KrgBased(SurrogateModel):
                 )
                 return res
 
-        limit, _rhobeg = 15 * len(self.options["theta0"]), 0.5
+        limit, _rhobeg = 10 * len(self.options["theta0"]) + 5, 0.5
         exit_function = False
         if "KPLSK" in self.name:
             n_iter = 1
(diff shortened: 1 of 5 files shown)


smt/utils/kriging.py
```diff
@@ -7,21 +7,20 @@
...
 
 from pyDOE2 import bbdesign
 from sklearn.metrics.pairwise import check_pairwise_arrays
 from smt.utils.design_space import BaseDesignSpace, CategoricalVariable
 
-try:
-    from numba import njit, prange
 
-    USE_NUMBA_JIT = True  # Set False to temporarily disable
+USE_NUMBA_JIT = int(os.getenv("USE_NUMBA_JIT", 0)) > 1e-9
```
relf, line 18: Why not just `USE_NUMBA_JIT = int(os.getenv("USE_NUMBA_JIT", 0))` ?

smt/utils/kriging.py
```diff
@@ -7,21 +7,20 @@
...
+prange = range
+if USE_NUMBA_JIT:
+    from numba import njit, prange
 
-except ImportError:  # pip install smt[numba]
-    USE_NUMBA_JIT = False
-    prange = range
+# Set False to temporarily disable
```
relf, line 23: to be removed

smt/surrogate_models/krg_based.py
```diff
@@ -1513,7 +1513,7 @@ def grad_minus_reduced_likelihood_function(log10t):
                 )
                 return res
 
-        limit, _rhobeg = 15 * len(self.options["theta0"]), 0.5
+        limit, _rhobeg = 10 * len(self.options["theta0"]) + 5, 0.5
```
relf, line 1516: `+5` required? Maybe `max(10 * len(self.options["theta0"]), 15)`

relf (changes requested):
Please, provide a description of your pull requests to explain the changes. It is useful  for users when we reference the PR and for future maintainance as well.

PR #456, opened 2023-07-18
Fix mixed hierarchical sampling

Paul-Saves:
Fix a bug over options for Mixed Hierarchical Sampling 
 Fix a rounding bug in  Mixed Hierarchical Sampling
Fix a sampling bug in EGO

### DIFF
--- a/smt/applications/ego.py
+++ b/smt/applications/ego.py
@@ -266,19 +266,14 @@ class EGO(SurrogateBasedApplication):
                 work_in_folded_space=True,
             )
             self._sampling = self.mixint.build_sampling_method(
-                LHS,
-                criterion="ese",
                 random_state=self.options["random_state"],
-                use_new_sampler=True,
             )
 
         else:
             self.mixint = None
             self._sampling = lambda n: self.design_space.sample_valid_x(
                 n,
-                criterion="ese",
                 random_state=self.options["random_state"],
-                use_new_sampler=True,
             )[0]
             self.categorical_kernel = None
 
--- a/smt/applications/ego.py
+++ b/smt/applications/ego.py
@@ -269,7 +269,7 @@ class EGO(SurrogateBasedApplication):
                 LHS,
                 criterion="ese",
                 random_state=self.options["random_state"],
-                new_sampler=True,
+                use_new_sampler=True,
             )
 
         else:
@@ -278,7 +278,7 @@ class EGO(SurrogateBasedApplication):
                 n,
                 criterion="ese",
                 random_state=self.options["random_state"],
-                new_sampler=True,
+                use_new_sampler=True,
             )[0]
             self.categorical_kernel = None
 
--- a/smt/applications/ego.py
+++ b/smt/applications/ego.py
@@ -21,6 +21,7 @@ from smt.utils.design_space import (
     FloatVariable,
     CategoricalVariable,
 )
+from smt.sampling_methods import LHS
 
 
 class Evaluator(object):
@@ -264,11 +265,21 @@ class EGO(SurrogateBasedApplication):
                 self.design_space,
                 work_in_folded_space=True,
             )
-            self._sampling = self.mixint.build_sampling_method()
+            self._sampling = self.mixint.build_sampling_method(
+                LHS,
+                criterion="ese",
+                random_state=self.options["random_state"],
+                new_sampler=True,
+            )
 
         else:
             self.mixint = None
-            self._sampling = lambda n: self.design_space.sample_valid_x(n)[0]
+            self._sampling = lambda n: self.design_space.sample_valid_x(
+                n,
+                criterion="ese",
+                random_state=self.options["random_state"],
+                new_sampler=True,
+            )[0]
             self.categorical_kernel = None
 
         # Build DOE
--- a/smt/utils/design_space.py
+++ b/smt/utils/design_space.py
@@ -748,9 +748,9 @@ class DesignSpace(BaseDesignSpace):
 
         # Simplified implementation: sample design vectors in unfolded space
         x_limits_unfolded = self.get_unfolded_num_bounds()
-        if "random_state" in kwargs.keys():
+        if "random_state" in kwargs:
             self.seed = kwargs["random_state"]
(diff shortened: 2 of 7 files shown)


smt/utils/test/test_design_space.py
```diff
@@ -176,7 +176,7 @@ def test_design_space(self):
         self.assertEqual(len(ds.design_variables), 4)
         self.assertTrue(np.all(~ds.is_conditionally_acting))
 
-        ds.sample_valid_x(3)
+        ds.sample_valid_x(3, random_state=42)
```
relf, line 179: Make me think the new option `new_sampler` is not tested.

smt/utils/design_space.py
```diff
@@ -299,7 +299,9 @@ def _decode_dv(x_encoded: np.ndarray, i_dv_decode):
         ]
         return decoded_des_vectors[0] if is_1d else decoded_des_vectors
 
-    def sample_valid_x(self, n: int, unfolded=False) -> Tuple[np.ndarray, np.ndarray]:
+    def sample_valid_x(
```
relf, line 302: Why not having `random_state` and `new_sampler` as arguments with default values? `kwargs` is less explicit and there is no documentation for the new options here.

relf (changes requested):

smt/utils/design_space.py
```diff
@@ -738,17 +743,25 @@ def _correct_get_acting(self, x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
         # Simplified implementation: sample design vectors in unfolded space
         x_limits_unfolded = self.get_unfolded_num_bounds()
-        sampler = LHS(xlimits=x_limits_unfolded, random_state=self.seed)
-        x = sampler(n)
-
-        # Cast to discrete and fold
-        self._normalize_x(x)
+        if "random_state" in kwargs.keys():
+            self.seed = kwargs["random_state"]
+        if "new_sampler" in kwargs.keys() and kwargs["new_sampler"]:
```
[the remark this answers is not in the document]
Paul-Saves: Yes ! Really good idea !

smt/utils/test/test_design_space.py
relf: Yes for EGO, butr it is a feature of `sample_valid_x` and to keep `DesignSpace` well-tested as it was, it would be nice to have here some specific tests of  `use_new_sampler`. Remember also we do not always run all EGO tests.

smt/utils/design_space.py
relf: It was designed for EGO but still is an option and a feature of a `DesignSpace` method. Why not `random_state=None, use_new_sampler=False` as defaults?

Finally, I am not  sure to understand why `use_new_sampler` is required by EGO. Why using the same LHS sampler would be bad?

smt/utils/design_space.py
```diff
@@ -738,17 +743,25 @@ def _correct_get_acting(self, x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
-        sampler = LHS(xlimits=x_limits_unfolded, random_state=self.seed)
-        x = sampler(n)
-
-        # Cast to discrete and fold
-        self._normalize_x(x)
+        if "random_state" in kwargs:
+            self.seed = kwargs["random_state"]
+        if kwargs.get("use_new_sampler"):
+            kwargs.pop("use_new_sampler", None)
+            if self.use_new_sampler:
```
relf, line 755: The logic here is a bit convoluted. Why do you need `self.use_new_sampler`? Another point for testing properly `use_new_sampler` in isolation in `DesignSpace` tests. ;)

relf (changes requested):

smt/utils/design_space.py
```diff
@@ -299,7 +299,9 @@ def _decode_dv(x_encoded: np.ndarray, i_dv_decode):
         ]
         return decoded_des_vectors[0] if is_1d else decoded_des_vectors
 
-    def sample_valid_x(self, n: int, unfolded=False) -> Tuple[np.ndarray, np.ndarray]:
+    def sample_valid_x(
+        self, n: int, unfolded=False, random_state=None, **kwargs
```
relf, line 303: Is `kwargs` really needed? Do we not just use LHS with `criterion="ese"` all the time? Do we need such flexibility at the moment?

Otherwise we should document what is expected in `kwargs`, eg options for the inner LHS sampler

smt/utils/design_space.py
```diff
@@ -738,17 +744,23 @@ def _correct_get_acting(self, x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
+    def _sample_valid_x(
+        self, n: int, random_state=None, **kwargs
+    ) -> Tuple[np.ndarray, np.ndarray]:
         """Sample design vectors"""
 
         # Simplified implementation: sample design vectors in unfolded space
         x_limits_unfolded = self.get_unfolded_num_bounds()
-        sampler = LHS(xlimits=x_limits_unfolded, random_state=self.seed)
-        x = sampler(n)
+        self.seed = random_state
```
relf, line 754: Why creating `self.seed`? Why not just use `random_state`? One call to `sample_valid_x` with `random_state`, will alter behaviour of further calls without `random_state` which seems error-prone to me.

relf (changes requested):

smt/utils/design_space.py
Paul-Saves: Do you want to fix the option to be "ese" and remove kwargs ?

relf: Yes! If there is no compelling need, let's simplify!

smt/applications/ego.py
```diff
@@ -264,11 +265,17 @@ def _setup_optimizer(self, fun):
                 self.design_space,
                 work_in_folded_space=True,
             )
-            self._sampling = self.mixint.build_sampling_method()
+            self._sampling = self.mixint.build_sampling_method(
+                LHS,
```
relf, line 269: `LHS` is useless here. See other comment.

relf (approved):

PR #467, opened 2023-10-06
add MFK and MFKPLS compatibility with mixed variables

RemyCharayron:
add MFK and MFKPLS compatibility with mixed variables using CR kernel.


smt/applications/mfk.py
```diff
@@ -483,25 +653,71 @@ def predict_variances_all_levels(self, X):
...
         C = self.optimal_par[0]["C"]
 
         beta = self.optimal_par[0]["beta"]
         Ft = solve_triangular(C, F, lower=True)
-        # yt = solve_triangular(C, self.y_norma_all[0], lower=True)
-        r_ = self._correlation_types[self.options["corr"]](
-            self.optimal_theta[0], d
-        ).reshape(n_eval, self.nt_all[0])
+
+        if not (self.is_continuous):
```
relf, line 672: See previous comment

relf (changes requested):
Some minor changes to be done. As it works only for CR, it would be nice to add a test in `check_params` and raise an exception if another kernel is selected. 

We could have subclassed MFK to avoid the occurences of `if self.continuous... else...` at the same time it is easier to read the algorithm in just one file. We'll keep this for a further refactoring if needs be.

smt/applications/mfk.py
```diff
@@ -958,25 +958,27 @@ def _check_param(self):
...
-            )
 
-            n_param = compute_n_param(
-                self.design_space,
-                self.options["categorical_kernel"],
-                d,
-                n_comp,
-                mat_dim,
-            )
+        if self.options["categorical_kernel"] is not None:
```
relf, line 962: This test can be removed

RemyCharayron: If we use MFK with continuous variables only, we will not set self.options["categorical_kernel"]. Then we will have self.options["categorical_kernel"] = None.  In this case removing the test "if self.options["categorical_kernel"] is not None:" will lead to raise the error "Only the continuous relaxation kernel is available with multi-fidelity" whereas we don't need to set a categorical kernel since only continuous variables are considered.

relf: Oups miss the `else`! Ok!

PR #504, opened 2024-01-29
Update to 2024 version of black for reformatting

Paul-Saves:


Paul-Saves:
Update for: 
- Black 24
- Scikit-learn 1.24 (not available on python3.8)
- Numpy 1.25 (not available on python3.8)

- Scikit-learn "manhattan distance" with "sum over features"  has been deprecated in Sklearn 1.24.

PR #506, opened 2024-01-31
Clean mixed integer sampling & seeding

Paul-Saves:
- Uniformize usage between design_space.sample_valid_x() andMixedIntegersampling _call__() 
- Separate RandomState & seed parameters : RandomState by default

### DIFF
--- a/smt/applications/mixed_integer.py
+++ b/smt/applications/mixed_integer.py
@@ -45,7 +45,7 @@ class MixedIntegerSamplingMethod(SamplingMethod):
         self._design_space = design_space
         if "random_state" in kwargs:
             self._design_space.seed = kwargs["random_state"]
-        else:
+        elif self._design_space.seed is None:
             self._design_space.seed = 42
         self._unfolded_xlimits = design_space.get_unfolded_num_bounds()
         self._output_in_folded_space = kwargs.get("output_in_folded_space", True)
--- a/smt/applications/mixed_integer.py
+++ b/smt/applications/mixed_integer.py
@@ -44,9 +44,9 @@ class MixedIntegerSamplingMethod(SamplingMethod):
         )
         self._design_space = design_space
         if "random_state" in kwargs:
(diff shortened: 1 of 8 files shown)


smt/applications/mixed_integer.py
```diff
@@ -43,6 +43,10 @@ def __init__(self, sampling_method_class, design_space, **kwargs):
             category=DeprecationWarning,
         )
         self._design_space = design_space
+        if "random_state" in kwargs:
+            self._design_space.seed = kwargs["random_state"]
```
relf, line 47: Hmm... AFAIK `seed` is an integer while `random_state` may be an `np.random.RandomState` or an integer. We should check the type here. Maybe the naming seed/random_state is not consistent throughout SMT. We could also start using more type hints.

smt/utils/design_space.py
```diff
@@ -742,6 +744,11 @@ def _is_num(val):
                     cs_vars[name] = CategoricalHyperparameter(name, choices=dv.values)
                 else:
                     raise ValueError(f"Unknown variable type: {dv!r}")
+            seed = None
```
relf, line 747: I suggest to make a utility function `to_seed()` as these lines are repeated twice later on.

smt/utils/design_space.py
```diff
@@ -927,11 +934,18 @@ def _sample_valid_x(
...
         # Simplified implementation: sample design vectors in unfolded space
         x_limits_unfolded = self.get_unfolded_num_bounds()
-        if self.seed is None:
-            self.seed = random_state
+        if self.random_state is None:
+            self.random_state = random_state
 
         if self._cs is not None:
             # Sample Configuration objects
+            if not (hasattr(self, "seed")):
```
relf, line 942: Testing the presence of an attribute is brittle. We should discuss to see if we can avoid that.

relf (changes requested):

relf (approved):

PR #508, opened 2024-02-05
Add Compound Symmetry kernel

Paul-Saves:
- Add Compound Symmetry kernel to the list of mixed integer kernels


relf (commented):
I suppose we need to update the documentation. Maybe you can commit in this PR the relevant .rstx, .rst files.

PR #509, opened 2024-02-09
fix bug for evaluating noise automatically for redundants points with noisy output when the inputs are mixed categorical

Paul-Saves:


smt/surrogate_models/krg_based.py
```diff
@@ -435,6 +435,18 @@ def _new_train(self):
         if not self.options["eval_noise"]:
             self.optimal_noise = np.array(self.options["noise0"])
         elif self.options["use_het_noise"]:
+            X = self.training_points[None][0][0]
```
relf, line 438: `X` and `y` seems to be already set at the beginning of the function and the standardization seems to be done just above. I understand there are side-effects on X above but It is quite difficult to understand the flow between mixed-variables stuff and noise management. Maybe you should add comments to explain what is going on with X and y. And how mixed-variables and noise management are related. Does it make sense to have noise on a discrete variable?

Paul-Saves: The noise is either global (noise0) and can be estimated automatically (eval_noise) or it is point-wise (hétéroscedastic) and given.

relf: Ok, so my question is why do you set X and y again and standardize again? Seems like mixed-variable handling above has an impact here otherwise this is already done.

Paul-Saves: Sure, I'll update on that to change the standardization position and rename variables that are different!

relf (approved):

PR #517, opened 2024-02-21
Jenn

shb84:
### Summary 

This PR replaces the `neural_net` package by [`jenn`](https://pypi.org/project/jenn/) as an upstream dependency, which is an updated version that is about 5x faster.  The previous implementation `genn` inadvertently copied data which has been fixed in `jenn` by updating arrays in place. 

#### Major changes: 

- Added `jenn>=1.0.1` as upstream dependency in `requirements.txt`
- Removed `stm.surrogate_models.neural_net`
- Using `jenn.model.NeuralNet` instead of `smt.utils.neural_net.model.Model`

#### Minor changes: 

- Removed: 
    - _`genn.load_smt_data` function replaced by `genn.GENN.load_data` method_
    - _`stm.examples.multi_modal` b/c it used API of underlying `GENN` model rather than SMT API (confusing to users)_
    - _`plot_training_history` and `goodness_of_fit` because they did not conform to SMT API_
    - _`genn.run_example` because it duplicates the unit test / usage example in docs_ 
    - _`smt.tests.test_all.genn` function b/c it wasn't really needed and stood out relative to the other surrogates_
    - _`@unittest.skipIf(int(os.getenv("RUN_SLOW", 0)) < 1, "too slow")` b/c `jenn` should fix issue with slowness_
    - _`smt.utils.neural_net` since it is replaced by `jenn`
    - _GENN tolerance relaxations in `test_all.py`, except for `exp` which could only be reduced to 10%_
- Changed: 
    - _Replaced `deep`, `wide` by `hidden_layer_sizes` (`GENN` class options)_ 
    - _Added `is_normalize` (`GENN` class options)_ 
    - _Added `is_bracktracking` (`GENN` class options)_ 
    - _`genn` docs sections (adjusted for `GENN` class modifications)_
    - _`learning_airfoil_parameters` example (adjusted for `GENN` class modifications)_
- Added: 
    - _`jenn` as required package in `setup.py`_
    - _(suggestion) `environment.yml` for development convenience `conda env update -f environment.yml -n smt`_


relf:
Hi. I pulled your PR locally and run pytest (after installing `jenn`), I got an error on the import, see below

```python
>python
Python 3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from jenn.model import NeuralNet
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'jenn.model'
>>>       
```

Then I noticed that `pip install -U jenn` installs `jenn 0.1.0` while I see there is a `jenn 1.0.0`.
So I tried:

```
>pip install jenn==1.0.0
ERROR: Ignored the following versions that require a different python version: 1.0.0 Requires-Python >=3.11
ERROR: Could not find a version that satisfies the requirement jenn==1.0.0 (from versions: 0.0.8, 0.1.0)
ERROR: No matching distribution found for jenn==1.0.0
```
As for now SMT is tested on py3.9, py3.10, py3.11, so...

shb84:
Thank you, Rémy. That makes sense. I did require Python 11 or greater on jenn, as I hadn't yet tested it on lower versions, which would explain. I will relax that requirement, do some testing, and circle back. This is very helpful. Thanks again.

shb84:
Ready for review.

relf (approved):
Hi Steven. This looks good to me, thank you!

PR #519, opened 2024-02-22
Remove some warnings

relf:
This PR deals with some warnings raised with: 
* Python 3.8
* numpy 1.23.4, 
* scipy 1.10.1, 
* scikit-learn 1.3.1, 
* numba 0.57.0
* ConfigSpace 0.6.1

CI tests will present warnings triggered by more recent versions of those dependencies

### DIFF
--- a/smt/surrogate_models/tests/test_kpls.py
+++ b/smt/surrogate_models/tests/test_kpls.py
@@ -46,25 +46,6 @@ class TestKPLS(unittest.TestCase):
             kriging.set_training_values(x, y)
             self.assertRaises(ValueError, lambda: kriging.train())
 
-    def test_kpls_training_with_zeroed_outputs(self):
-        # Test scikit-learn 0.24 regression cf. https://github.com/SMTorg/smt/issues/274
-        for corr_str in [
-            "pow_exp",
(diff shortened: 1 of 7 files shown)


smt/surrogate_models/tests/test_kpls.py
Paul-Saves, line 1: Why removing this test ?

Paul-Saves (changes requested):

smt/surrogate_models/tests/test_kpls.py
```diff
@@ -46,25 +46,6 @@ def test_predict_output(self):
             kriging.set_training_values(x, y)
             self.assertRaises(ValueError, lambda: kriging.train())
 
-    def test_kpls_training_with_zeroed_outputs(self):
```
Paul-Saves, line 49: Why removing this test?

relf: It was a test for a regression in scikit-learn 0.24 on a ill-formed case which now just generates a warning, as KPLS is tested elsewhere, I just removed the test. I can bring it back if you a find a value in it.

Issue #571, opened 2024-05-22
SIGBUS (Misaligned Address Error)


NAThompson:
To reproduce:

```
$ python3
Python 3.11.9
$ uname -a
Darwin  23.4.0 Darwin Kernel Version 23.4.0: ; root:xnu-10063.101.17~1/RELEASE_ARM64_T6020 arm64
$ ./rmtc_issue.py
fish: Job 1, './rmtc_issue.py' terminated by signal SIGBUS (Misaligned address error)
```

Code:

```python
#!/usr/bin/env python3
from math import pi as π
import random
import numpy as np
from smt.surrogate_models import RMTB, RMTC
import matplotlib.pyplot as plt
import num_dual


dimension = 4
def generate_sine_surface_data(samples=10):
    training_points = np.full(shape=(samples, dimension), fill_value=np.nan)
    training_values = np.full(shape=samples, fill_value=np.nan)

    for i in range(samples):
        x = np.random.uniform(-1, 1, dimension)
        training_points[i, :] = x
        v = 1
        for j in range(dimension):
            v *= np.sin(π*x[j])
        training_values[i] = v
    return training_points, training_values


def rmtc_surrogate_model(training_points, training_values):
    limits = np.full(shape=(dimension, 2), fill_value=np.nan)
    for i in range(dimension):
        limits[i, 0] = -1
        limits[i, 1] = 1

    model = RMTC(
        xlimits=limits,
        num_elements=20,
        energy_weight=1e-15,
        regularization_weight=0.0,
        print_global=False,
    )
    model.set_training_values(training_points, training_values)
    model.train()
    return model


if __name__ == "__main__":
    training_points, training_values = generate_sine_surface_data()
    rmtc = rmtc_surrogate_model(training_points, training_values)
```

PR #572, opened 2024-05-23
Fix signed integer overflow

NAThompson:
In Issue #571, I encountered a bus error.
Using AddressSanitizer, I traced the error to a negative memory index caused by a 32 bit signed integer overflowing the bounds and requesting memory from a negative index.

Swap out `int` for `long` so that more total memory can be addressed and allow larger problem sizes.


relf:
Thanks!

PR #581, opened 2024-05-29
Document exponential dependence of runtime on training points for KPLS

NAThompson:


relf (changes requested):
Thanks for the benchmarking. We certainly should strive to monitor algorithms runtime performances. That said if you are interested by a faster KPLS, you should try [GPX](https://smt.readthedocs.io/en/stable/_src_docs/surrogate_models/gpx.html) (it is a wrapping of a rewrite of the KRG/KPLS core algorithms in Rust, cf. [egobox](https://pypi.org/project/egobox/) library). You just have to change the model creation:

```python
model = GPX(theta0=[1e-2], kpls_dim=1)
```

Regarding the documentation contribution, you should take a look [here](https://smt.readthedocs.io/en/stable/_src_docs/dev_docs.html#building-the-documentation), .rst files are generated from .rstx files. If you change the .rst manually you should ensure the .rstx is in sync otherwise changes might be lost if the whole doc is regenerated and committed.
You can also checkout the doc generated from your PR [here](https://smt--581.org.readthedocs.build/en/581/_src_docs/surrogate_models/kpls.html). The image should be resized smaller or else removed as the exponential complexity is somehow already highlighted in this caveat section.

relf:
On the algorithm complexity, I should have mentioned that KRG/KPLS/KPLSK/GPX belongs to gaussian process regression family which are indeed O(n^3) in the number of training points. 

If you have a large number of training points you might want to try our [sparse gaussian process implementation (SGP)](https://smt.readthedocs.io/en/stable/_src_docs/surrogate_models/sgp.html) which adresses this issue.

relf:
Documentation was updated in #591 to give an overview of gaussian process surrogates

PR #582, opened 2024-06-06
Add spatial derivatives to squar_sin_exp kernel

NicolasJeanGonel:
Note that in order to make this derivative work a dummy variable is used, "dd", and passed to the kernel while being 0 and not be used. This is because squar_sin_exp has 2 times more hyperparameter than the other kernels, so some code has to be bypassed.

Some tests with finite differences have been relaxed, in order to account for the possibility of having a finite difference of 0 and a real derivative that is a very small number (giving a relative error of infinity), see the formula of np.testing.assert_allclose for more details.


smt/surrogate_models/tests/test_krg_predictions.py
```diff
@@ -24,10 +24,17 @@ def pb(x):
...
+            # sin + linear trend
+            y = np.sin(x[:, 1]) + 2 * x[:, 1] + 5 * x[:, 0] + 10  # + linear trend
+            return y
+
         xlimits = np.array([[-5, 10], [-5, 10]])
         self.sampling = LHS(xlimits=xlimits, random_state=42)
         self.xt = self.sampling(12)
         self.yt = pb(self.xt)
+        self.yt_squar_sin_exp = pb_for_sin_squar_exp(self.xt)
+        # def pb_for_sinsquar
```
relf, line 37: To be removed

relf (changes requested):
Thanks for this contribution. A minor cleanup and I merge.

PR #586, opened 2024-06-14
Added 4 tests for noisy GPs and for GPX

MaelTremouille:
Added 4 tests to complete the tests in the test_ego.py file and added design_space features in GPX

### DIFF
--- a/smt/applications/ego.py
+++ b/smt/applications/ego.py
@@ -71,13 +71,6 @@ class EGO(SurrogateBasedApplication):
             types=int,
             desc="Maximum number of internal optimizations",
         )
-        # New declaration for noise0
-        declare(
-            "noise0",
-            [0.0],
-            types=(list, np.ndarray),
-            desc="Initial noise hyperparameters",
-        )
         declare("n_start", 20, types=int, desc="Number of optimization start points")
         declare(
             "n_parallel",
@@ -116,7 +109,7 @@ class EGO(SurrogateBasedApplication):
         declare(
             "surrogate",
             KRG(print_global=False),
-            types=(KRG, KPLS, KPLSK, GEKPLS, MGP, GPX), #####
+            types=(KRG, KPLS, KPLSK, GEKPLS, MGP, GPX),
             desc="SMT kriging-based surrogate model used internaly",
         )
         self.options.declare(
--- a/smt/applications/ego.py
+++ b/smt/applications/ego.py
@@ -17,7 +17,7 @@ from smt.applications.mixed_integer import (
     MixedIntegerSamplingMethod,
 )
 from smt.sampling_methods import LHS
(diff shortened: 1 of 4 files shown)


smt/applications/ego.py
```diff
@@ -71,6 +71,13 @@ def _initialize(self):
             types=int,
             desc="Maximum number of internal optimizations",
         )
+        # New declaration for noise0
```
relf, line 74: Remove useless comment (changes are visible in the PR).

smt/applications/ego.py
```diff
@@ -71,6 +71,13 @@ def _initialize(self):
...
+        declare(
+            "noise0",
```
relf, line 76: I do not see any usage of options["noise0"]. We do not need to declare a `noise0` option in EGO, it is already an option of `KRG`. To be removed.

smt/applications/tests/test_ego.py
```diff
@@ -87,6 +88,48 @@ def test_function_test_1d(self):
...
+        xlimits = np.array([[0.0, 25.0]])
+        criterion = "EI"
+        design_space = DesignSpace(xlimits)
+        noise0 = [1e-1] # fixed variance for 1D
+
+        ego = EGO(
+            n_iter=n_iter,
+            criterion=criterion,
+            n_doe=3,
+            surrogate=KRG(design_space=design_space, print_global=False),
```
relf, line 123: `noise0` should be passed in KRG call

smt/applications/tests/test_ego.py
```diff
@@ -87,6 +88,48 @@ def test_function_test_1d(self):
...
+            surrogate=surrogate,
+            random_state=42,
+        )
+
+        x_opt, y_opt, _, _, _ = ego.optimize(fun=TestEGO.function_test_1d)
+
+        self.assertAlmostEqual(18.9, x_opt.item(), delta=1)
+        self.assertAlmostEqual(-15.1, y_opt.item(), delta=1)
+        
+    def test_function_ego_GPN_1D(self):
```
relf, line 112: maybe use `noisy_KRG` instead of `GPN`

relf (changes requested):
Thanks for your contribution! You should install the VScode extension Ruff and configure to automatically format the code when you save files, otherwise use `ruff format <file>`.

MaelTremouille:
Review of the comments from the last PR and corrections

MaelTremouille:
I formated my files with Ruff

relf (changes requested):
You have to create another branch restarting from master containing the changes related to the feature 'no theta optimization'. Moreover, instead of adding a new option maybe we can use the `hyper_opt` option and add a `NoOptim` value or even better `None` to state that `theta0` is used as is.

PR #591, opened 2024-06-18
Add a GPR section in the documentation

relf:
This PR adds a section dedicated to gaussian process regression surrogates (aka kriging). It allows to give an overview of the various gaussian process surrogate flavours.


relf:
The new section can be previewed [here](https://smt--591.org.readthedocs.build/en/591/_src_docs/surrogate_models/gpr.html)

PR #592, opened 2024-06-18
Noisy gp and gpx tests

MaelTremouille:
This PR aims to implement four new tests: two tests to check if the noisy GP works correctly in 1D and 2D, and two tests to check if GPX works correctly in 1D and 2D.


MaelTremouille:
I truly do not understand the error because the test runs perfectly for me.

smt/applications/tests/test_ego.py
```diff
@@ -152,6 +194,48 @@ def test_rosenbrock_2D_SBO(self):
         self.assertTrue(np.allclose([[1, 1]], x_opt, atol=1))
         self.assertAlmostEqual(0.0, y_opt.item(), delta=1)
 
+    @unittest.skipIf(int(os.getenv("RUN_SLOW_TESTS", 0)) < 1, "too slow")
```
relf, line 197: Removes this guard, we do not know if the test is slow for now

smt/applications/tests/test_ego.py
```diff
@@ -152,6 +194,48 @@ def test_rosenbrock_2D_SBO(self):
...
+    def test_function_ego_GPX_rosenbrock_2d(self):
```
relf, line 198: To be consistent with previous test names
```suggestion
    def test_rosenbrock_2D_GPX(self):
```

relf (changes requested):

relf:
Tests fail due to new scipy 1.14 release which removes some old deprecated features we are still using... To be fixed

Issue #585, opened 2024-06-14
Relevance of the warning `R is too ill conditioned...`


relf:
When running the first example of kriging in the tutorial (reproduced here)
```python
from smt.utils.misc import compute_rms_error
from smt.problems import Rosenbrock
from smt.sampling_methods import LHS
from smt.surrogate_models import KRG

ndim = 2
ndoe = 20  # int(10*ndim)
# Define the function
fun = Rosenbrock(ndim=ndim)

# Construction of the DOE
# in order to have the always same LHS points, random_state=1
sampling = LHS(xlimits=fun.xlimits, criterion="ese", random_state=1)
xt = sampling(ndoe)
# Compute the outputs
yt = fun(xt)

# Construction of the validation points
ntest = 200  # 500
sampling = LHS(xlimits=fun.xlimits, criterion="ese", random_state=1)
xtest = sampling(ntest)
ytest = fun(xtest)

# The variable 'theta0' is a list of length ndim.
t = KRG(theta0=[1e-2] * ndim, print_prediction=False)
t.set_training_values(xt, yt[:, 0])

t.train()

# Prediction of the validation points
y = t.predict_values(xtest)
print("Kriging,  err: " + str(compute_rms_error(t, xtest, ytest)))
```
I get the following output

```
python test_warning.py
___________________________________________________________________________

                                  Kriging
___________________________________________________________________________

 Problem size

      # training points.        : 20

___________________________________________________________________________

 Training

   Training ...
D:\rlafage\workspace\smt\smt\surrogate_models\krg_based.py:968: UserWarning: R is too ill conditioned. Poor combination of regression model and observations.
  warnings.warn(
   Training - done. Time (sec):  7.4901881
Kriging,  err: 0.00010937725926990396
```

While the error is pretty small we have this warning telling us something is wrong which is a pretty confusing.

Paul-Saves:
Yes, I have to be more restrictive on this test

Paul-Saves:
t.check_training_numerically() : 

Minimal distance between two points in any direction is 5.84e-04
Correlation matrix R condition number is 3.07e+13
None

Paul-Saves:
With pow_exp: Minimal distance between two points in any direction is 8.48e-04
Correlation matrix R condition number is 8.48e+02

PR #602, opened 2024-06-25
Add sgp notebook, fix typo and fix warning 

Paul-Saves:
- Fix https://github.com/SMTorg/smt/issues/585
- Add sgp analytical examples notebook
- Fix a typo in doc of mfk
- MGP was not tested properly within test_all.py


smt/surrogate_models/krg_based.py
```diff
@@ -220,7 +220,7 @@ def _final_initialize(self):
 
         # initialize default power values
         if self.options["corr"] == "squar_exp":
-            self.options["pow_exp_power"] = 2.0
+            self.options["pow_exp_power"] = 2.0 - 1e-8
```
relf, line 223: Hmmm... Sorry but **squared** exponential exponent is equal to 2!  ;-) I guess you try to handle a numerical issue but it cannot be done like that.

PR #606, opened 2024-06-27
Implementation of re-interpolated sigma2 to improve the performance of the EI criterion

MaelTremouille:
The goal is to perform a fictitious re-interpolation to reduce the variance when using the EI criterion with noisy GPs. We ensure that points are fictitiously re-interpolated when using the criterion to avoid selecting infill points that we already have. Added a test to verify that the variance using re-interpolation is lower than without.


smt/surrogate_models/krg_based.py
```diff
@@ -136,6 +136,12 @@ def _initialize(self):
         # theta in (0,1e1], is a good choice to avoid overfitting. By standardising
         # X in R, X_norm = (X-X_mean)/X_std, then X_norm in [-1,1] if considering
         # one std intervals. This leads to theta in (0,2e1]
+        declare(
```
relf, line 139: Should not be in this PR (as the related code below). Moreover we can avoid to declare another option by specifying a `None` value for the `hyper_opt` option.

relf (changes requested):
Please keep the PR focused on the re-interpolation for EGO only. The theta fixed feature should be in another dedicated PR

smt/surrogate_models/krg_based.py
MaelTremouille: I apologize, I forgot that it was in branch master

smt/surrogate_models/krg_based.py
```diff
@@ -954,18 +955,21 @@ def _reduced_likelihood_function(self, theta):
             r = self._correlation_types[self.options["corr"]](theta, self.D).reshape(
                 -1, 1
             )
-        R = np.eye(self.nt) * (1.0 + nugget + noise)
+        R = np.eye(self.nt) * (1.0 + nugget)
+        R_noisy = np.eye(self.nt) * (1.0 + nugget + noise)
         R[self.ij[:, 0], self.ij[:, 1]] = r[:, 0]
```
relf, line 960: Why not pass `r` to _compute_sigma2_ri and set `R` only if needed ?

smt/surrogate_models/krg_based.py
```diff
@@ -1006,7 +1010,18 @@ def _reduced_likelihood_function(self, theta):
...
             sigma2.sum()
-        ) - self.nt * np.log10(detR)
+        ) - self.nt * np.log10(detR_noisy)
+
+        eval_noise = self.options["eval_noise"]
+        is_noise0 = self.options["noise0"] != [0.0]
+        # noisy KRG
+        sigma2_ri = None
+        is_noisy = eval_noise or is_noise0
+        sigma2_ri = None
```
relf, line 1020: duplicated. To be removed.

smt/surrogate_models/krg_based.py
```diff
@@ -1006,7 +1010,18 @@ def _reduced_likelihood_function(self, theta):
...
+        if is_noisy:
```
relf, line 1021: Prefer more compact:  `if self.options["eval_noise"] or self.options["noise0"] != [0.0]:   # is_noisy` instead of the five lines above and you can assign `par["sigma2_ri"]` directly

relf (changes requested):

smt/surrogate_models/krg_based.py
MaelTremouille: fixed

Paul-Saves:
Il faut refactorer pour utiliser la même fonction pour _compute_sigma2 et _compute_sigma2_ri

PR #607, opened 2024-06-28
Add Mixed integer KPLSK

Paul-Saves:
- Add Mixed integer KPLSK for CR & GD kernels

@Paul-Saves
[update KPLSK mixed: krg_based & test. Add a boolean for testing second loop.


relf (changes requested):
Please undo the 'ruff' commit, ruff formatting when touching numeroius files has to be done in a separate dedicated PR.

Paul-Saves:
Ruff in https://github.com/SMTorg/smt/pull/608

Issue #612, opened 2024-07-02
Documentation hyperlink failing


NicolasJeanGonel:
The link in the documentation [Handling of Mixed Integer variables](https://github.com/SMTorg/smt/blob/master/tutorial/SMT_MixedInteger_application.ipynb), should return a notebook but is giving a 404 error.
 It can be found [here](https://smt.readthedocs.io/en/latest/_src_docs/getting_started.html). It is entitled "Handling of mixed integer variables".

relf:
fixed by  #617

PR #617, opened 2024-07-08
Fix issue #612, broken link in the documentation

NicolasJeanGonel:
This is a fix to the issue #612. The file had been renamed, so now the hyperlink is renamed too.


relf:
Thanks!

PR #618, opened 2024-07-08
fix kplsk for retraining 

Paul-Saves:
Mixed integer KPLSK models can not be retrained, that is, just changing training values and train.
This problem may occurs  for optimization when applying some clustering techniques, for example in SEGOMOE


relf (commented):
The code in krg_based should be refactored. Having a bunch of booleans (`kplsk_second_loop`, `exit_function`, ...) which trigger various behaviours in a unique function is very brittle and looks very much like a footgun. At least for KPLSK code (which should live in kplsk file) we should find a better way.

PR #622, opened 2024-07-10
Class based kernels for kriging

NicolasJeanGonel:
Refactoring of the kernels to make them class based.
 The body of the functions(class) is still the same, but this should allow easier further development and manipulations of the kernels.
Theta is now an internal variable of the kernel class, instead of a variable for the __call__ function. This causes it to change a lot but allows tracking of its value.


smt/surrogate_models/mgp.py
```diff
@@ -346,11 +342,13 @@ def _predict_variance_derivatives_hyper(self, x, u=None):
...
 
         # Compute the correlation function
-        r = (
-            self._correlation_types[self.options["corr"]](theta, d, d_x=d_x)
-            .reshape(n_eval, self.nt)
-            .T
-        )
+        self.corr.theta = theta
+        r = self.corr(d, d_x=d_x).reshape(n_eval, self.nt).T
+        # r = (
```
Paul-Saves, line 347: You may remove the commented previous code that has been already updated

smt/surrogate_models/mgp.py
```diff
@@ -382,13 +380,14 @@ def _predict_variance_derivatives_hyper(self, x, u=None):
...
         for omega in range(len(self.optimal_theta)):
-            drdomega = (
-                self._correlation_types[self.options["corr"]](
-                    theta, d, grad_ind=omega, d_x=d_x
-                )
-                .reshape(n_eval, self.nt)
-                .T
-            )
+            drdomega = self.corr(d, grad_ind=omega, d_x=d_x).reshape(n_eval, self.nt).T
+            # drdomega = (
```
Paul-Saves, line 384: To remove

Paul-Saves (changes requested):

relf (approved):
Thanks for this refactoring! It will definitely improve the maintainance of SMT and pave the way for future kernel composition. As soon as the requested minor changes are commited, I will merge.

PR #626, opened 2024-07-17
Clean utils after the switch from genn to jenn (thanks to Steven Berguin)

Paul-Saves:
Removed not used anymore files. Notwithstanding, those files were not working anymore.


Paul-Saves:
Once again, thank you @shb84 !

PR #627, opened 2024-07-18
Update design space utils 

Paul-Saves:


Paul-Saves:
@relf The PR is now ready !

relf (commented):
What is the difference between `ConfigSpace`and `ArchDesignSpaceGraph` ? Should we keep both ? I am starting to think that all this stuff should be maintained in a another repo as SMT exposes `DesignSpace` interface to be isolated from specific implementations and should not depend on external librairies for this.

PR #631, opened 2024-07-25
SGP feature in EGO

MaelTremouille:
This PR aims to implement SGP in EGO. Each time, we add the new point to the training data and keep the same induced data to train the model. It also allows for creating the set of induced points using the k-means method.


MaelTremouille:
Issue with EI computation and minimization

PR #646, opened 2024-09-16
 Implementation of EGO with noisy data by @MaelTremouille

Paul-Saves:
[MaelTremouille](https://github.com/MaelTremouille) commented [Jul 22, 2024]

The goal is to perform a fictitious re-interpolation to reduce the variance when using the EI criterion with noisy GPs. We ensure that points are fictitiously re-interpolated when using the criterion to avoid selecting infill points that we already have. Added a test to verify that the variance using re-interpolation is lower than without.


relf:
supersede #630

PR #647, opened 2024-09-16
fix CR and hier, twice corrected

Paul-Saves:
The correction was done twice for CR. Hence the GP model was not interpolating when using CR for decreed categorical variables.


smt/applications/tests/test_mixed_integer.py
```diff
@@ -1057,10 +1059,10 @@ def test_hierarchical_design_space_example_all_categorical_decreed(self):
             sm.train()
             y_s = sm.predict_values(Xt)[:, 0]
             _pred_RMSE = np.linalg.norm(y_s - Yt) / len(Yt)
-
+            self.assertTrue(_pred_RMSE < 1e-7)
```
relf, line 1062: You can use `assertLessThan` which gives more descriptive error message.

relf (approved):

PR #651, opened 2024-09-17
Move advanced design space handling in separate `smt-design-space-ext` module

Paul-Saves:
Create the connection between SMT and SMT-Design-Space. 
Remove code related to SMT-Design-Space moved in a new repo 

--- SMTOrg/SMT-Design-Space
- Add the class ArchDesignSpaceGraph(DesignSpace) to DesignSpace.py
- Improve seeding in DesignSpace.py
- Fix a bug when rounding and the lower bound is not zero
----
- Improve tests & notebook accordingly


Paul-Saves:
@relf Okay for me!

smt/design_space/__init__.py
```diff
@@ -0,0 +1,55 @@
+import importlib
+
+spec = importlib.util.find_spec("smt_design_space")
+if spec:
+    HAS_SMTDesignSpace = True
```
relf, line 5: constant are all uppercase ie HAS_SMT_DESIGN_SPACE. I would prefer HAS_SMT_DESIGN_SPACE_EXT or HAS_DESIGN_SPACE_EXT for a shorter name.

Paul-Saves:
I fixed the initialization for Collab.

relf (changes requested):
Actually the module name `smt_design_space` is confusing, I suggest to rename it in `smt_design_space_ext` for external or extension I mean smt has an API for design space and a standalone internal basic design space handling which can be enhanced using `smt_design_space_ext` module.
Sorry for this late suggestion. [Naming is a well-know hard problem in computer science]( https://martinfowler.com/bliki/TwoHardThings.html)

smt/design_space/tests/test_design_space.py
```diff
@@ -8,17 +8,36 @@
...
-    BaseDesignSpace,
-    CategoricalVariable,
-    DesignSpace,
-    FloatVariable,
-    IntegerVariable,
-    OrdinalVariable,
-)
+import os
+
+if os.getenv("FORCE_RUN_PLAIN_BASICS"):
```
relf, line 14: Would not it be the opposite? I mean when the variable is set we want to run the "else" statements? Maybe the variable should be renamed RUN_PLAIN_DESIGN_SPACE_TEST as it is only used for test purpose.

relf (approved):

PR #664, opened 2024-10-17
 fix CONT_RELAX with Hierarchical variables

Paul-Saves:


relf (approved):
Anything to add?

Paul-Saves:
It seems okay for me

PR #684, opened 2024-12-02
Remove zeroing small variance values

### LINKED ISSUE #681: Bug with the function predict_variances
Hello,

I have a bug with the function predict_variances.  It seems like variances are rounded to 0 when they are too small.
A code producing the bug is attached. A simple function with mixed variables is used, a DoE is performed and surrogate models are built. 10 points' variances are predicted and displayed. The bug appears when the DoE size is big enough (20 points are sufficient) and disappears when the size is small (5 points for instance). The bug appears from the version 2.5.1 (it wasn't here in 2.5.0).

I have solved it by commenting the line 1859 from the file smt/smt/surrogate_models/krg_based.py

Thanks !
 
[bug_version_SMT.zip](https://github.com/user-attachments/files/17932897/bug_version_SMT.zip)

Paul-Saves:
Fix #681 . Thanks to L. Baraton


smt/surrogate_models/krg_based.py
```diff
@@ -1856,7 +1856,6 @@ def _predict_variances(
             A = self.optimal_par["sigma2"]
         B = 1.0 - (rt**2.0).sum(axis=0) + (u**2.0).sum(axis=0)
         # machine precision: force to zero!
```
relf, line 1858: You have to remove the comment too!

PR #690, opened 2024-12-10
Dev castano

mcastanoUQ:
Adding anisotropic kernel computation, and regularization term to the likelihood.


relf:
You have to rebase your branch on current SMT master or create a new branch from current SMT master, otherwise you have your old commits in the way.

PR #692, opened 2024-12-11
Anisotropic kernel for the multi-fidelity co-Kriging model (MFCK)

mcastanoUQ:
Adding anisotropic kernel computation for the multi-fidelity co-Kriging approach proposed (MFCK) to deal with non-nested DoEs—additional modification adding a regularization parameter in the Negative Marginal Likelihood.


relf:
Please update title and description to reflect the PR content.

mcastanoUQ:
It's not necessary to have those values. I'll update the PR with the default values of Cobyla

relf (changes requested):
After the commented code removal, I will merge if you have nothing to add.

PR #698, opened 2024-12-16
Add metric computation functions for surrogate performance evaluation 

Antoine-Averland:


smt/utils/misc.py
```diff
@@ -63,7 +63,7 @@ def standardization(X, y):
     return X, y, X_offset, y_mean, X_scale, y_std
 
 
-def compute_rms_error(sm, xe=None, ye=None, kx=None):
+def compute_relative_error(sm, xe=None, ye=None, kx=None):
```
relf, line 66: Yes this is a relative error not an RMSE

smt/utils/misc.py
```diff
@@ -108,6 +108,35 @@ def compute_rms_error(sm, xe=None, ye=None, kx=None):
...
+    N = len(ye)
+    ye2 = sm.predict_values(xe)
+    variance = sm.predict_variances(xe)
+
+    error = (ye2 - ye) ** 2 / variance
+    pva = np.sum(error) / N
+    return pva
+
+
+def compute_rmse_error(sm, xe, ye):
```
relf, line 122: This function name reads like "compute root mean squared error error". I suggest to remove "_error" from the functions names: `compute_pva`, `compute_rmse` and `compute_q2`

relf (changes requested):

PR #726, opened 2025-04-23
Add the activation vector as an input of hierarchical based functions

Paul-Saves:


smt/problems/problem.py
```diff
@@ -83,7 +83,12 @@ def sample(self, n):
         x, _ = self.design_space.sample_valid_x(n)
         return x
 
-    def __call__(self, x: np.ndarray, kx: Optional[int] = None) -> np.ndarray:
+    def __call__(
+        self,
+        x: np.ndarray,
+        delta: Optional[np.ndarray] = None,
```
relf, line 89: Why `delta` name while we pass `self.eval_is_acting` below. I would also prefer to add this new optional parameter at the end (it won't break legacy code).

smt/problems/problem.py
```diff
@@ -116,8 +121,10 @@ def __call__(self, x: np.ndarray, kx: Optional[int] = None) -> np.ndarray:
         # Correct the design vector and get information about which design variables are active
         x_corr, self.eval_is_acting = self.design_space.correct_get_acting(x)
         self.eval_x = x_corr
-
-        y = self._evaluate(x_corr, kx)
+        if bool(np.max(self.design_space.is_conditionally_acting)):
```
relf, line 124: Do you mean `if np.any(self.design_space.is_conditionally_acting)` ?

smt/problems/tests/test_problem_examples.py
```diff
@@ -80,6 +80,30 @@ def test_hier_neural_network(self):
         plt.ylabel("y")
         plt.show()
 
+    @unittest.skipIf(NO_MATPLOTLIB, "Matplotlib not installed")
+    def test_hier_goldstein(self):
```
relf, line 84: This code is intended to be in the documentation. It is just run to check if there is no runtime error. But currently it is not used. Besides there is no test (I mean no test with assertions) for this new parameter you introduce while you change the SMT API.

relf (changes requested):
Hi Paul. You change the `Problem` class call method which is one of the main API of SMT. [Documentation has to be updated](https://smt.readthedocs.io/en/stable/_src_docs/dev_docs/dev_problems.html) accordingly to explain the new parameter. 
As a problem developper, what should I do with this parameter? What is the impact on the result? When should I use it?

I understand you change the API to manage hierarchical problems in SMT. At the moment we have only two problems of this type which are [not even shown in the documentation](https://smt.readthedocs.io/en/stable/_src_docs/problems.html). Do we expect to add other benchmark problems of this type? 

Moreover I feel that hierarchical and differentiable problems are a bit orthogonal. It makes me wonder if we should start with another base class like `HierarchicalProblem` (maybe I am going too far here).

smt/problems/problem.py
```diff
@@ -135,6 +152,11 @@ def _evaluate(self, x: np.ndarray, kx: Optional[int] = None) -> np.ndarray:
         kx : int or None
             Index of derivative (0-based) to return values with respect to.
             None means return function value rather than derivative.
+        eval_is_acting : ndarray of shape (n, nx) or (n,), optional
+            Boolean mask indicating for each evaluation point which design
+            variables are “active” (i.e.\ truly participating) in a
+            hierarchical design space.  If provided, `evaluate` can use this
+            to skip or correct inactive dimensions.  Default is! None.
```
relf, line 159: is! ?

smt/problems/branin.py
```diff
@@ -8,10 +8,19 @@
 
 import numpy as np
 
+from smt.design_space import DesignSpace
 from smt.problems.problem import Problem
 
 
 class Branin(Problem):
+    @property
```
relf, line 16: You've added this for all problems, it should be in the base class `Problem`

smt/tests/test_problems.py
```diff
@@ -72,16 +72,16 @@ def run_test(self, problem):
...
                 abs_rms_error_FD = np.linalg.norm(dydx_FD - dydx_AN)
                 rel_rms_error_FD = np.linalg.norm(dydx_FD - dydx_AN) / np.linalg.norm(
                     dydx_FD
                 )
 
                 abs_rms_error_CS = np.linalg.norm(dydx_CS - dydx_AN)
-                rel_rms_error_CS = np.linalg.norm(dydx_CS - dydx_AN) / np.linalg.norm(
-                    dydx_CS
+                rel_rms_error_CS = np.linalg.norm(dydx_CS - dydx_AN) / (
+                    np.linalg.norm(dydx_CS) + 1e-18
```
relf, line 84: Consider using `np.finfo(float).eps` instead of 1e-18

relf (changes requested):

PR #740, opened 2025-07-09
Add a wrapper for scikit-learn

Paul-Saves:


smt/surrogate_models/mgp.py
```diff
@@ -199,10 +199,10 @@ def predict_variances_no_uq(self, x):
 
     def _check_xdim(self, x):
         _, n_features = x.shape
-        nx = self.nx
+        nx = np.copy(self.nx)
```
relf, line 202: `np.copy` useless as `self.nx` is an integer

PR #809, opened 2026-04-27
Implement specialized predict methods in MFCK/SMFCK

mcastanoUQ:
Addition of wrapper functions to predict the conditional mean and conditional variance in MFCK.

- predict_values()
- predict_variances_all_levels()
- predict_variances()

The function now returns the predictions in a numpy array instead of returning them in a list (this change keeps consistency with MFK). Useful for benchmarks where you want to compare the predictive performance with (MFK, SMFK).

Update of the notebooks of SMFK, MFCK, and SMFCK with the last update; the notebooks had some problems regarding the predictive mean.

-In SMFCK is now necessary to define "theta0" exactly as in MFK. You need to define it in the same dimensional space of your inputs (before, you just needed to put an initial value, and MFCK automatically assigned the same initial theta to all the dimensions)

-As SMFK integrates sparse approximations for some experiments is necessary to increase the nugget (near 1e-10) to guarantee the inversion of the matrix.


smt/applications/mfck.py
```diff
@@ -549,13 +549,241 @@ def predict_values(self, x, is_acting=None):
...
+                        )
+                    )
+
+        beta1 = solve_triangular(L, np.vstack(k_xX), lower=True)
+
+        alpha1 = solve_triangular(L, self.y_norma_all, lower=True)
+
+        return self.y_std * np.dot(beta1.T, alpha1) + self.y_mean
+
+    def predict_variances(self, x: np.ndarray) -> np.ndarray:
```
relf: Keep the original function signature even if you do not use the arguments
```
    def predict_variances(
        self, X: np.ndarray, is_acting=None, is_ri=False
    ) -> np.ndarray:
```

smt/applications/mfck.py
```diff
@@ -568,10 +796,130 @@ def predict_variances(
...
+                if self.options["use_het_noise"]:
+                    k_xx = self.compute_diag_K(x, x, ind, ind, self.optimal_theta)
+                else:
+                    k_xx = self.compute_diag_K(
+                        x, x, ind, ind, self.optimal_theta[: -self.lvl]
+                    )
+            else:
+                k_xx = self.compute_diag_K(x, x, ind, ind, self.optimal_theta)
+
+            for j in range(self.lvl):
```
relf: I think there is code duplication. Please create relevant helper methods to avoid repeating the same code

smt/applications/mfck.py
```diff
@@ -568,10 +796,130 @@ def predict_variances(
         -------
         y : np.ndarray
             Evaluation point output variable values
+        Theta : Optimal parameters.
```
relf: It does not return `sigma_rhos` like mfk or smfk. And why return `self.theta_optimal` which can be accessed.

relf (changes requested):
You should explain what you have fixed in the code. Please, do not just state "update" I can see that, it does not provide any information. 

The title and description of the PR should focus on the source code fix, not notebooks updates (only in description). Why did you rewrite predict methods? Why did you change arguments and return types?

smt/applications/mfck.py
mcastanoUQ: I was confused here, the comments of MFK say that the functions just return "y", but when I checked at the end, it was returning something else, the regression of the Rho values scaled with sigma square. As I'm not interested in analysing that, I just returned the optimal values of the optimal theta of MFCK. I can just return "y" if it's better for the understanding of the code

smt/applications/mfck.py
mcastanoUQ: Hi @relf, the function returns only the predictive variance across all fidelity levels. Similar to MFK, I used the same formulas as in predict_all_levels(), but in this case, it's just the prediction of the conditional variance.

smt/applications/mfck.py
mcastanoUQ: Changed, ready for a PR.

smt/applications/mfck.py
relf: My point here is that this for loop appears three times in the code you add, please just put this code chunk in a new method and just call it where it is relevant.

smt/applications/mfck.py
relf: Ok just return "y"

relf:
I do not understand why so many changes if the point was just to translate lists into numpy arrays.

PR #816, opened 2026-06-09
Add the possibility to give an array-like input for n_comp option in KPLS

enricostragiotti:
… and test the range of n_dim and choose the best one. It has been refactored to take into account the advice given by @Paul-Saves.

I added unit tests (some of them are quite long due to the exhaustive search, so I marked them as `@unittest.skipIf(int(os.getenv("RUN_SLOW_TESTS", 0)) < 1, "too slow")`).

I updated the docs with a new example and added some description to the parameters.

I also added a little safeguard on the number of k_fold min based on the number of training points:
```python
k_fold = min(4, original_nt)
```

### DIFF
--- a/smt/surrogate_models/kpls.py
+++ b/smt/surrogate_models/kpls.py
@@ -21,7 +21,6 @@ class KPLS(KrgBased):
     def _initialize(self):
         super(KPLS, self)._initialize()
         declare = self.options.declare
-
         declare(
             "n_comp",
             1,
@@ -165,11 +164,11 @@ class KPLS(KrgBased):
 
         Parameters
         ----------
-        X : ndarray, shape (original_nt, n_features) — full training inputs, never mutated
-        y : ndarray, shape (original_nt, n_outputs)  — full training outputs, never mutated
-        n_comp : int — number of PLS components to use
-        k_fold : int — number of folds
-        original_nt : int — size of the full dataset; used to compute fold width
+        X : ndarray, shape (original_nt, n_features), full training inputs, never mutated
+        y : ndarray, shape (original_nt, n_outputs), full training outputs, never mutated
+        n_comp : int, number of PLS components to use
+        k_fold : int, number of folds
+        original_nt : int, size of the full dataset; used to compute fold width
                       and the PRESS normalisation factor. Passed explicitly so
                       this method does not depend on self.nt, which may have
                       been overwritten by a previous fold.
@@ -255,6 +254,8 @@ class KPLS(KrgBased):
                 if error < best_error:
                     best_error = error
                     best_comp = n_comp
+        else:
+            raise ValueError("Strategy '" + strategy + "' not implemented.")
 
         # Restore training data and state
         self.training_points[None][0][0] = X
--- a/smt/surrogate_models/kpls.py
+++ b/smt/surrogate_models/kpls.py
@@ -160,6 +160,11 @@ class KPLS(KrgBased):
         nbk = int(self.nt / k_fold)
         press_m = []
         n_comp_min, n_comp_max = map(int, self.options["n_comp"])  # We want int only
+        if n_comp_min >= n_comp_max:
+            raise ValueError(
+                "n_comp must be an integer or an array-like of length 2 "
+                "[min_n_comp, max_n_comp], with min_n_comp < max_n_comp"
+            )
         n_comp_values = list(range(n_comp_min, n_comp_max + 1))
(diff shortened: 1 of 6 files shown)


smt/surrogate_models/kpls.py
```diff
@@ -21,7 +21,12 @@ def _use_pls(self) -> bool:
     def _initialize(self):
         super(KPLS, self)._initialize()
         declare = self.options.declare
-        declare("n_comp", 1, types=int, desc="Number of principal components")
+        declare(
+            "n_comp",
+            1,
+            types=(list, tuple, np.ndarray, int),
+            desc="Number of principal components",
+        )
```
Paul-Saves, line 32: In our current design, passing a list to n_comp implies that a method needs multiple parameters—such as performing multiple dimension reductions for categorical variables. Let's find a way to implement this great feature without overloading the n_comp argument.

Paul-Saves (commented):
What is the difference between this implementation and  KPLS_auto? 

Also, could we update the code to keep the original type constraints for n_comp? Since it represents a specific component count, I really think it needs to remain an int or numpy.integer. I really like the addition of an external function like _estimate_number_of_components_from_range for testing, but I’d prefer we don't alter the core input types for n_comp.

enricostragiotti:
> What is the difference between this implementation and KPLS_auto?
> 
> Also, could we update the code to keep the original type constraints for n_comp? Since it represents a specific component count, I really think it needs to remain an int or numpy.integer. I really like the addition of an external function like _estimate_number_of_components_from_range for testing, but I’d prefer we don't alter the core input types for n_comp.

The need of testing a full range of n_comp comes only from the fact that we observed that the error has not always a  constant concavity.

For example here is what I obtain if I write the press_m error

```
def test_sphere_KPLS(self):
        self.run_test()
```
with ndim=10 (as you did in the unit tests of the auto kpls.

```
[np.float64(2678.3722874967852),
np.float64(2624.683060968535), 
np.float64(2633.044933733534), 
np.float64(2597.5496823546437), 
np.float64(2539.413749676094), 
np.float64(2697.310960760146), 
np.float64(2780.8667615569043), 
np.float64(2799.2309127682674), 
np.float64(2852.136549502163), 
np.float64(2853.358912903915)]
```

As you can see, even if the error for n_comp=3 is greater than the error for n_comp=2 (2 is then the best for the auto algorithm, as written in the actual unit test in `test_kpls_auto.py`), once we eval n_comp=4 and 5 we see that the error decrease again, with the best being 5 in this specific case. I'm definitely not an expert in this, and I think that maybe we can have auto converge to 5 if we modify the k fold or the n_start, but I still see the interest in having the possibility to test a whole range.

Maybe the auto algorithm is improvable by adding some kind of inertia in the choice of n_comp? And thus not stopping at the first point that increase the error?

smt/surrogate_models/kpls.py
enricostragiotti: This is what we come up with @relf and @NatOnera, but I'm more than open to change it! I'm in general not a big fan on the API changes neither :)

Paul-Saves:
I completely agree with you, stopping as soon as possible was proposed by Wold to avoid too many computations. We don't want to try everything every time, but adding the possibility to test more values is indeed beneficial. It would be nice if your code and the existing one could be merge, though, as they serve the same purpose 🤔 
```
# Conceptual example of the merged logic
def estimate_number_of_components(X, y, max_comp=10, strategy='wold', k_fold=None):
    """
    Estimates the optimal number of components for KPLS.
    
    Parameters:
    - strategy: 'wold' (stops at first error increase) or 'exhaustive' (tests all to max_comp)
    """
    best_comp = 1
    best_error = float('inf')
    errors = []

    for n in range(1, max_comp + 1):
        error = calculate_press_error(X, y, n_comp=n, k_fold=k_fold)
        errors.append(error)
        
        # Original KPLS_auto logic (Wold's early stopping)
        if strategy == 'wold':
            if error > best_error:
                break # Stop searching, we hit the local minimum
            best_error = error
            best_comp = n
            
        # Your new logic (Exhaustive search)
        elif strategy == 'exhaustive':
            if error < best_error:
                best_error = error
                best_comp = n

    return best_comp
```

Paul-Saves:
The PR seems ready, everything's good to me!

relf (approved):
LGTM! Just for the record, we will have to refactor the code to avoid changing `self.options` which should be immutable (strictly user inputs) and use a dedicated `self.n_comp` attribute.

PR #821, opened 2026-06-23
Adding RBFGen to SMT

spcvanschie:
This is to add the recently proposed surrogate modeling method "RBFGen", from the research group of @hwangjt, to SMT. RBFGen works in two steps:

1. First, we construct an underconstrained RBF basis (with more RBFs than data points) and compute a least-squares solution to the interpolation problem.
2. Then, we compute the nullspace of the above RBF basis and construct a neural network that maps from a set of latent variables to the nullspace. The neural network is trained based on knowledge of the model output that is not used in the above interpolation problem: This can include things such as output monotonicity, positivity or known output distributions at specific input points.

As a result, we get a surrogate model that incorporates expert knowledge into the surrogate modeling process without affecting the interpolation of the given data points. This approach is covered in detail in [this paper](https://asmedigitalcollection.asme.org/computingengineering/article/26/7/071005/1230105/Knowledge-Guided-Generative-Surrogate-Modeling-for).

On the software side: 

- This adds three files to SMT: The surrogate model itself is contained in ```surrogate_models/rbfgen.py```, which inherits from the utility files ```utils/nn_rich_rbf.py``` and ```utils/nn_lossterms.py```.
- Training the neural network requires the use of PyTorch. Scikit-learn does not have the necessary methods to handle this. We've included PyTorch as an optional dependency, in the same way that EGObox is handled.
- Step 1 above requires control over the RBF basis. Therefore we've included a separate RBF class named ```NNRichRBF``` in ```utils/nn_rich_rbf.py```, rather than using the RBF model that is already contained in SMT.
- We have included a few loss terms for training the neural network. These all inherit from the same ```LossTerm``` base class in ```utils/nn_lossterms.py```, which allows users to define their own context-specific loss terms.


Paul-Saves:
Hi @spcvanschie , thanks a lot for this contribution and for the detailed write-up — RBFGen looks like an interesting addition, and the way you've structured it (NNRichRBF, modular loss terms) is clean.

One question on the dependency side: I noticed this adds PyTorch as a new dependency. 
Since SMT has historically relied on scikit-learn, could you help me understand whether PyTorch is essential for this approach, or if there's a lighter-weight alternative? Just want to make sure we keep the install footprint manageable for users.
.Happy to discuss further — thanks again for the work here!

relf:
@Paul-Saves, unfortunately, they did not succeed to make it work with scikit-learn (see second bullet above) hence the introduction of the **optional** dependency torch only for the most motivated... and provided we succeed to install it! 😉

requirements.txt
```diff
@@ -13,3 +13,4 @@ ruff            # format and lint code
 setuptools-scm  # version from git tags
 jenn >= 2.0.0, <3.0
 egobox >= 0.36, <1.0
+pytorch >= 2.4.1
```
relf: I think you want `torch` not `pytorch`

```suggestion
torch >= 2.4.1
```

spcvanschie:
Hi @Paul-Saves, I agree that adding PyTorch as a new (optional) dependency is not ideal. We looked into using Scikit-learn instead, but it doesn't allow us to use the same clean approach that PyTorch gives us access to. Because of this, we decided to import PyTorch as an optional dependency, in the same way that EGObox is imported.

requirements.txt
spcvanschie: Yes good catch, I've pushed this change to the repository.

relf:
@spcvanschie you have to format code with ruff 0.15

smt/surrogate_models/rbfgen.py
```diff
@@ -0,0 +1,217 @@
...
+        self.supports["variances"] = True
+
+        self.generator = None
+        self.opt_generator = None
+        self.network_weights = None
+
+        self.loss_terms = []
+
+        # We'll need to store training data because add_... methods use it
+        self.x_train = None
```
relf: Why `self.x_train` and `self.y_train` when there is `self.training_points` ?

smt/surrogate_models/rbfgen.py
```diff
@@ -0,0 +1,217 @@
...
+
+        self.supports["variances"] = True
+
+        self.generator = None
+        self.opt_generator = None
+        self.network_weights = None
+
+        self.loss_terms = []
+
+        # We'll need to store training data because add_... methods use it
```
relf: There is just one method "add_..." method,  `add_loss_term()` ?

relf (commented):
Thanks for your contribution, LGTM! Some minor questions below

relf (changes requested):
You have to add some tests with assertions. See `smt/tests/test_all.py` and under `smt/surrogate_models/tests`

smt/surrogate_models/rbfgen.py
spcvanschie: Great point, this was an accidental leftover from a past version. I've removed these class properties since they aren't being used anywhere, and I've scanned the rest of the files for similar bugs.

smt/surrogate_models/rbfgen.py
spcvanschie: Yes, we previously used an approach where each loss term had its own "add_..." method. That turned out to be cumbersome, so to streamline things we refactored the code into its current form. I forgot to remove this line of comments in the process. I've fixed that now.

smt/tests/test_all.py
```diff
@@ -37,6 +37,12 @@
 except ImportError:
     COMPILED_AVAILABLE = False
 
+try:
```
relf: This is for C++ compiled surrogates it does not work for RBFGen. You have to mimick what is done for GPX above

relf (changes requested):
You need to rebase your branch on master, run ruff format and fix the import (see comment below)

smt/tests/test_all.py
spcvanschie: Ok! I've updated this to mimic the way GPX is being handled.

relf (approved):
Thank you for this contribution!

PR #829, opened 2026-08-28
Fix correlation matrix conditionning check

### LINKED ISSUE #824: Missing transpose in is_training_ill_conditioned method?
I see in krg_based.py the following method

```
  def is_training_ill_conditioned(self):
        """
        Check if the training dataset could be an issue and print both
        the dataset correlation matrix condition number and
        minimal distance between two points.
        ----
        Returns true if R is ill_conditionned
        """
        R = self.optimal_par["C"] @ self.optimal_par["C"]
```

Since `C` comes from a Cholesky factorization of the matrix `R`, shouldn't it be the case that `R = self.optimal_par["C"] @ self.optimal_par["C"].T`? That is $R=CC^{\top}$. Let me know if I'm mistaken and any comments to help clear up my confusion would, as always, be appreciated.

relf:
Fixes #824 

I've changed the checking to just use the cond number 
as I am not sure why we check the smallest eigenvalue (maybe due to the bug?).
I've also reworked the test to be simpler.


thartland:
Great, thanks for this!

PR #830, opened 2026-09-09
Update in MFCK/ SMFCK

mcastanoUQ:
**Summary**

Extends the MFCK and SMFCK multi-fidelity surrogate models with gradient-based
hyperparameter optimization, heteroscedastic noise prediction, and a sequential
optimization scheme. Also includes an internal refactoring that makes the shared
co-Kriging machinery reusable and inspectable.

**Modularization**

Shared logic previously duplicated across the MFCK and SMFCK implementations has
been extracted into helper methods:

- Cholesky decomposition of the covariance matrix
- Cross-covariance assembly between fidelity levels
- Diagonal prior covariance construction

Beyond removing duplication, this exposes the intermediate quantities so that
internal behaviour (conditioning, cross-level correlations) can be inspected
directly rather than only through the final prediction.

**Gradients for MFCK**

Analytical gradients of the log marginal likelihood with respect to the kernel
hyperparameters are now implemented for the RBF kernel. This allows MFCK to be
trained with gradient-based optimizers such as TNC.

**Heteroscedastic noise prediction**

Both MFCK and SMFCK can now predict input-dependent noise

**Sequential optimization**
An option that integrates hyperparameter optimization level by level rather
than jointly. This option reduces the training complexity of MFCK to make it comparable to that of MFK.

## Notebooks

Updated to use the sequential optimization option in all tutorials associated:

- `tutorial/MultiFi/SMT_MFCK_tutorial.ipynb`
- `tutorial/MultiFi/SMT_SMFCK_tutorial.ipynb`
- `tutorial/MultiFi/SMT_SMFK_tutorial.ipynb`
- `tutorial/MultiFi/Paper_SMFCK_Notebook_Rosenbrock.ipynb`


smt/applications/smfk.py
```diff
@@ -396,9 +396,14 @@ def predict_variances_all_levels(self, X, is_acting=None):
                     + sigma2 * (u_**2).sum(axis=0)
                 )
             else:
+                print(
```
relf: print statement should be removed

smt/applications/smfk.py
```diff
@@ -396,9 +396,14 @@ def predict_variances_all_levels(self, X, is_acting=None):
...
+                    self.optimal_noise_all[i].shape,
+                    (r_t**2).sum(axis=0).shape,
+                    (u_**2).sum(axis=0).shape,
+                )
                 MSE[:, i] = sigma2 * (
                     1
-                    + self.optimal_noise_all[i]
+                    # + self.optimal_noise_all[i]
```
relf, line 401: to be removed or ... ?

relf (approved):
Thanks. Minor comments to address before merging

smt/applications/smfk.py
mcastanoUQ: Removing it

smt/applications/smfk.py
mcastanoUQ: Yes, now it presents an inconsistency with the shapes. I need to check that; for the moment, we can remove it. That line is to take into account the noise provided by the experts when working with het. noise

relf: Ok so just update the PR description to mention that smfk does not take into account the noise

